PR #3, opened 2015-02-23
Fix bug where deep matcher expects a map passed into the matches function

TedSander:
Get around a bug in core matchers where matchers expect a non-null map value for deep matcher. See: https://github.com/dart-lang/matcher/issues/1


fibulwinter:
Thanks, Ted

PR #8, opened 2015-05-11
Tweaks

kevmoo:


fibulwinter:
I agree with most of the changes except:
1. Why we need to explicitly override noSuchMethod?
```dart
class MockFoo extends AbstractFoo with Mock {
  noSuchMethod(i) => super.noSuchMethod(i);
}
```
instead of 
```dart
class MockedClass extends Mock implements RealClass {}
```

2. Explicit type annotation of argument for named(), verifyZeroInteractions and verifyNoMoreInteractions: IMHO it just makes tests harder to read as we need to add "as Mock" to please type-check.

matanlurey:
Explicitly overriding noSuchMethod with the @proxy annotation tells Dart analyzer you meant not to implement all methods, and not to hint/warn that methods are missing.

fibulwinter:
Ok, I have added this to c5add5d27c75b006ad7d860228671f04abe0c2ef.

PR #9, opened 2015-05-13
Travis config.

pq:
@fibulwinter @TedSander


pq:
@fibulwinter : see related issue #10 as well.  That bit needs to be flipped in order for builds to work.

PR #18, opened 2015-07-22
Added an explicit check of named arguments to InvocationMatcher to en…

lyceel:
…sure

that the match will fail if the specified named arguments in the actual
invocation differ from the mocked invocation.

Updated tests to include test cases for the above change.

Changed the top-level "any" and "captureAny" _ArgMatchers to use the
"anything" Matcher from test, rather than using null and having to test
for a null Matcher in isMatchingArg().

### DIFF
--- a/lib/mockito.dart
+++ b/lib/mockito.dart
@@ -149,6 +149,12 @@ class InvocationMatcher {
       }
       index++;
     }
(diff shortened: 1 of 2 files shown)


TedSander:
Thanks for the contribution.

Issue #15, opened 2015-06-30
Bug: Using verify with a parameter that is a Map or Iterable should be supported


matanlurey:
For example:

```
myService.update(cssClasses: ['a', 'b])
```

Should match

```
verify(myService.update(cssClasses: ['a', 'b']))
```

(Even though the `cssClasses` `List` instance is different)

TedSander:
So this is actually possible today but maybe the interface isn't as nice as it could be. This is how you can do it:
```dart
verify(myService.update(cssClasses: argThat(equals(['a', 'b']))))
```

argThat allows you to specify matchers. This is nice because then do can do shallow equals, unordered equals, contains, etc.

Here is the fall through code:
https://github.com/fibulwinter/dart-mockito/blob/master/lib/mockito.dart#L168

It looks like @fibulwinter was using identical at one time for the matcher, but then removed it. I'm not sure if there was a specific reason of moving away from the matcher. Maybe just for simplicity.

Probably easy enough to switch to equals matcher if there isn't a specific reason for the current pattern.

fibulwinter:
The root problem is https://github.com/dart-lang/sdk/issues/2217. To counter this I wanted to use equality matcher by default, but I vaguely remember that there were some complications... Stay tuned.

matanlurey:
Ack, the workaround is good enough for today, but I would like to use equals if it is OK.

@fibulwinter: It looks like dart-lang/sdk#2217 is long solved, we commonly use the `equals` matcher in practically every unit test without issue.

mockturtl:
+1 for defaulting to `equals`, with [`argThat(same(x))`](http://www.dartdocs.org/documentation/matcher/0.12.0%2B1/index.html#matcher/matcher@id_same) for when identity matters.

fibulwinter:
Done in 0.11.0 - `equals` matcher is used by default.

PR #19, opened 2015-07-31
Fix List args equality check

nfrancois:
This fix allow list as parameters. 
Without test list are not compared together and verify is failing


fibulwinter:
Done as part of 0.11.0
Note that it uses more generic approach: equal matcher by defult (see https://github.com/fibulwinter/dart-mockito/issues/15)

PR #26, opened 2016-06-16
Add strong mode-compliant 'typed' API

### LINKED ISSUE #25: Allow `any` to work with Strong mode
This is a sibling issue to https://github.com/dart-lang/sdk/issues/26036.

The issue is that we cannot use `any` with Strong mode turned on. For example:

```dart
int f(List<int> l) => l[0];

void main() {
  f(any);
}
```

Yields an analyzer error (`dartanalyzer --strong a.dart`): "Unsound implicit cast from dynamic to List<int>".

I think we can solve this on the Mockito side, similar to Java Mockito's [`any()`](http://site.mockito.org/mockito/docs/current/org/mockito/Matchers.html#any()).

I have a prototype implementation that uses the Strong mode type parameter comments:

```dart
/*=T*/ any/*<T>*/([/*=T*/ e = null]) => null;
```

The tricky part is that `any` would now return `null`, instead of the `_ArgMatcher` (again similar to [Java Mockito's](https://github.com/mockito/mockito/blob/master/src/main/java/org/mockito/Matchers.java#L161..L163), and we would have to implement something like Java Mockito's [reportMatcher()](https://github.com/mockito/mockito/blob/master/src/main/java/org/mockito/AdditionalMatchers.java#L989..L991), maybe changing how the whole matching situation works...

What do you think? Any thoughts or other ideas?

srawlins:
Fixes #25 

CC @leafpetersen @kevmoo @TedSander 

This solution is modeled after Java Mockito's implementation. Unfortunately, Java doesn't have named args, so this implementation is a little uglier with regards to named args.

The README and the tests show usage. What do you think?

### Alternatives

@TedSander and I came up with the `typed/*<horror>*/(matcher, {name})` API. But we're not married to it, if people can think of something cleaner. Maybe a separate `typedNamed`?

We could not conceive of a way for the API to understand when an argument is passed as a named argument. We can't rely on order (`when(foo.fn(a: typed(...), b: typed(...)))`), because Invocation's namedArguments is a Map, possibly with unstable order.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,11 +126,11 @@ verifyNoMoreInteractions(cat);
 ```dart
 //simple capture
 cat.eatFood("Fish");
-expect(verify(cat.eatFood(capture)).captured.single, "Fish");
+expect(verify(cat.eatFood(captureAny)).captured.single, "Fish");
 //capture multiple calls
 cat.eatFood("Milk");
 cat.eatFood("Fish");
-expect(verify(cat.eatFood(capture)).captured, ["Milk", "Fish"]);
+expect(verify(cat.eatFood(captureAny)).captured, ["Milk", "Fish"]);
 //conditional capture
 cat.eatFood("Milk");
 cat.eatFood("Fish");
@@ -147,6 +147,91 @@ expect(cat.sound(), "Purr");
 //using real object
 expect(cat.lives, 9);   
 ```
+
+## Strong mode compliance
+
+Unfortunately, the use of the arg matchers in mock method calls (like `cat.eatFood(any)`)
+violates the [Strong mode] type system. Specifically, if the method signature of a mocked
+method has a parameter with a parameterized type (like `List<int>`), then passing `any` or
+`argThat` will result in a Strong mode warning:
+
+> [warning] Unsound implicit cast from dynamic to List&lt;int>
+
+In order to write Strong mode-compliant tests with Mockito, you might need to use `typed`,
+annotating it with a type parameter comment. Let's use a slightly different `Cat` class to
+show some examples:
+
+```dart
+class Cat {
+  bool eatFood(List<String> foods, [List<String> mixins]) => true;
+  int walk(List<String> places, {Map<String, String> gaits}) => 0;
+}
+
+class MockCat extends Mock implements Cat {}
+
+var cat = new MockCat();
+```
+
+OK, what if we try to stub using `any`:
+
+```dart
+when(cat.eatFood(any)).thenReturn(true);
+```
+
+Let's analyze this code:
+
+```
+$ dartanalyzer --strong test/cat_test.dart
+Analyzing [lib/cat_test.dart]...
+[warning] Unsound implicit cast from dynamic to List<String> (test/cat_test.dart, line 12, col 20)
+1 warning found.
+```
+
+This code is not Strong mode-compliant. Let's change it to use `typed`:
+
+```dart
+when(cat.eatFood(typed/*<List<String>>*/(any)))
+```
+
+```
+$ dartanalyzer --strong test/cat_test.dart
+Analyzing [lib/cat_test.dart]...
+No issues found
+```
+
+Great! A little ugly, but it works. Here are some more examples:
+
+```dart
+when(cat.eatFood(typed/*<List<String>>*/(any), typed/*<List<String>>*/(any)))
+    .thenReturn(true);
+when(cat.eatFood(typed/*<List<String>>*/(argThat(contains("fish")))))
+    .thenReturn(true);
+```
+
+Named args require one more component: `typed` needs to know what named argument it is
+being passed into:
+
+```dart
+when(cat.walk(
+    typed/*<List<String>>*/(any),
+    gaits: typed/*<Map<String, String>>*/(any), name: 'gaits')).thenReturn(true);
+```
+
+Note the `name` argument. Mockito should fail gracefully if you forget to name a `typed`
+call passed in as a named argument, or name the argument incorrectly.
+
+One more note about the `typed` API: you cannot mix `typed` arguments with `null`
+arguments:
+
+```dart
+when(cat.eatFood(null, typed/*<List<String>>*/(any))).thenReturn(true) // Throws!
+when(cat.eatFood(
+    argThat(equals(null)),
+    typed/*<List<String>>*/(any))).thenReturn(true); // Works.
+```
+
+[Strong mode]: https://github.com/dart-lang/dev_compiler/blob/master/STRONG_MODE.md
+
 ## How it works
 The basics of the `Mock` class are nothing special: It uses `noSuchMethod` to catch
 all method invocations, and returns the value that you have configured beforehand with
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ Analyzing [lib/cat_test.dart]...
 This code is not Strong mode-compliant. Let's change it to use `typed`:
 
 ```dart
-when(cat.eatFood(typed/*<List<String>>*/(any)))
+when(cat.eatFood(typed(any)))
 ```
 
 ```
@@ -202,32 +202,29 @@ No issues found
 Great! A little ugly, but it works. Here are some more examples:
 
 ```dart
-when(cat.eatFood(typed/*<List<String>>*/(any), typed/*<List<String>>*/(any)))
-    .thenReturn(true);
-when(cat.eatFood(typed/*<List<String>>*/(argThat(contains("fish")))))
-    .thenReturn(true);
+when(cat.eatFood(typed(any), typed(any))).thenReturn(true);
+when(cat.eatFood(typed(argThat(contains("fish"))))).thenReturn(true);
 ```
 
 Named args require one more component: `typed` needs to know what named argument it is
 being passed into:
 
 ```dart
-when(cat.walk(
-    typed/*<List<String>>*/(any),
-    gaits: typed/*<Map<String, String>>*/(any), name: 'gaits')).thenReturn(true);
+when(cat.walk(typed(any), gaits: typed(any, named: 'gaits')))
+    .thenReturn(true);
 ```
 
-Note the `name` argument. Mockito should fail gracefully if you forget to name a `typed`
+Note the `named` argument. Mockito should fail gracefully if you forget to name a `typed`
 call passed in as a named argument, or name the argument incorrectly.
 
 One more note about the `typed` API: you cannot mix `typed` arguments with `null`
 arguments:
 
 ```dart
-when(cat.eatFood(null, typed/*<List<String>>*/(any))).thenReturn(true) // Throws!
+when(cat.eatFood(null, typed(any))).thenReturn(true); // Throws!
 when(cat.eatFood(
     argThat(equals(null)),
-    typed/*<List<String>>*/(any))).thenReturn(true); // Works.
+    typed(any))).thenReturn(true); // Works.
 ```
 
 [Strong mode]: https://github.com/dart-lang/dev_compiler/blob/master/STRONG_MODE.md
--- a/lib/src/mock.dart
+++ b/lib/src/mock.dart
@@ -10,6 +10,8 @@ _WhenCall _whenCall = null;
 final List<_VerifyCall> _verifyCalls = <_VerifyCall>[];
 final _TimeStampProvider _timer = new _TimeStampProvider();
 final List _capturedArgs = [];
+final List<_ArgMatcher> _typedArgs = <_ArgMatcher>[];
+final Map<String, _ArgMatcher> _typedNamedArgs = <String, _ArgMatcher>{};
 
 // Hidden from the public API, used by spy.dart.
 void setDefaultResponse(Mock mock, dynamic defaultResponse) {
@@ -31,6 +33,9 @@ class Mock {
   }
 
   dynamic noSuchMethod(Invocation invocation) {
+    if (_typedArgs.isNotEmpty || _typedNamedArgs.isNotEmpty) {
+      invocation = _reconstituteInvocation(invocation);
+    }
     if (_whenInProgress) {
       _whenCall = new _WhenCall(this, invocation);
       return null;
@@ -53,6 +58,132 @@ class Mock {
       : identical(this, other);
 
   String toString() => _givenName != null ? _givenName : runtimeType.toString();
+}
+
+// Return a new [Invocation], reconstituted from [invocation], [_typedArgs],
+// and [_typedNamedArgs].
+Invocation _reconstituteInvocation(Invocation invocation) {
+  var newInvocation = new FakeInvocation(invocation);
+  return newInvocation;
+}
+
+/// An Invocation implementation that allows all attributes to be passed into
+/// the constructor.
+class FakeInvocation extends Invocation {
+  final Symbol memberName;
+  final Map<Symbol, dynamic> namedArguments;
+  final List<dynamic> positionalArguments;
+  final bool isGetter;
+  final bool isMethod;
+  final bool isSetter;
+
+  factory FakeInvocation(Invocation invocation) {
+    if (_typedArgs.isEmpty && _typedNamedArgs.isEmpty) {
+      throw new StateError("FakeInvocation called when no typed calls have been saved.");
+    }
+
+    // Handle named arguments first, so that we can provide useful errors for
+    // the various bad states. If all is well with the named arguments, then we
+    // can process the positional arguments, and resort to more general errors
+    // if the state is still bad.
+    var namedArguments = _reconstituteNamedArgs(invocation);
+    var positionalArguments = _reconstitutePositionalArgs(invocation);
+
+    _typedArgs.clear();
+    _typedNamedArgs.clear();
+
+    return new FakeInvocation._(
+        invocation.memberName,
+        positionalArguments,
+        namedArguments,
+        invocation.isGetter,
+        invocation.isMethod,
+        invocation.isSetter);
+  }
+
+  static Map<Symbol,dynamic> _reconstituteNamedArgs(Invocation invocation) {
+    var namedArguments = <Symbol, dynamic>{};
+    var _typedNamedArgSymbols = _typedNamedArgs.keys.map((name) => new Symbol(name));
+    invocation.namedArguments.forEach((name, arg) {
+      if (arg == null) {
+        if (!_typedNamedArgSymbols.contains(name)) {
+          // Incorrect usage of [typed], something like:
+          // `when(obj.fn(a: typed(any)))`.
+          throw new ArgumentError(
+              'A typed argument was passed in as a named argument named "$name", '
+              'but did not a value for its name. Each typed argument that is '
+              'passed as a named argument needs to specify the `name` argument. '
+              'For example: `when(obj.fn(x: typed(any, name: "x")))`.');
+        }
+      } else {
+        // Add each real named argument that was _not_ passed with [typed].
+        namedArguments[name] = arg;
+      }
+    });
+
+    _typedNamedArgs.forEach((name, arg) {
+      Symbol nameSymbol = new Symbol(name);
+      if (!invocation.namedArguments.containsKey(nameSymbol)) {
+        // Incorrect usage of [name], something like:
+        // `when(obj.fn(typed(any, name: 'a')))`.
+        throw new ArgumentError(
+            'A typed argument was declared with name $name, but was not passed '
+            'as an argument named $name.');
+      }
+      if (invocation.namedArguments[nameSymbol] != null) {
+        // Incorrect usage of [name], something like:
+        // `when(obj.fn(a: typed(any, name: 'b'), b: "string"))`.
+        throw new ArgumentError(
+            'A typed argument was declared with name $name, but a different '
+            'value (${invocation.namedArguments[nameSymbol]}) was passed as '
+            '$name.');
+      }
+      namedArguments[nameSymbol] = arg;
+    });
+
+    return namedArguments;
+  }
+
+  static List<dynamic> _reconstitutePositionalArgs(Invocation invocation) {
+    var positionalArguments = <dynamic>[];
+    var nullPositionalArguments =
+        invocation.positionalArguments.where((arg) => arg == null);
+    if (_typedArgs.length != nullPositionalArguments.length) {
+      throw new ArgumentError(
+          'null arguments are not allowed alongside typed(); use '
+          '"typed(eq(null))"');
+    }
+    int i = 0;
+    int j = 0;
+    while (i < _typedArgs.length && j < invocation.positionalArguments.length) {
+      var arg = _typedArgs[i];
+      if (invocation.positionalArguments[j] == null) {
+        // [typed] was used; add the [_ArgMatcher] given to [typed].
+        positionalArguments.add(arg);
+        i++;
+        j++;
+      } else {
+        // [typed] was not used; add the [_ArgMatcher] from [invocation].
+        positionalArguments.add(invocation.positionalArguments[j]);
+        j++;
+      }
+    }
+    while (j < invocation.positionalArguments.length) {
+      // Some trailing non-[typed] arguments.
+      positionalArguments.add(invocation.positionalArguments[j]);
+      j++;
+    }
+
+    return positionalArguments;
+  }
+
+  FakeInvocation._(
+      this.memberName,
+      this.positionalArguments,
+      this.namedArguments,
+      this.isGetter,
+      this.isMethod,
+      this.isSetter);
 }
 
 named(var mock, {String name, int hashCode}) => mock
@@ -300,6 +431,15 @@ get captureAny => new _ArgMatcher(anything, true);
 captureThat(Matcher matcher) => new _ArgMatcher(matcher, true);
 argThat(Matcher matcher) => new _ArgMatcher(matcher, false);
 
+/*=T*/ typed/*<T>*/(_ArgMatcher matcher, {String name}) {
+  if (name == null) {
+    _typedArgs.add(matcher);
+  } else {
+    _typedNamedArgs[name] = matcher;
+  }
+  return null;
+}
+
 class VerificationResult {
   List captured = [];
   int callCount;
@@ -412,4 +552,15 @@ void logInvocations(List<Mock> mocks) {
   allInvocations.forEach((inv) {
     print(inv.toString());
   });
+}
+
+/// Should only be used during Mockito testing.
+void resetMockitoState() {
+  _whenInProgress = false;
+  _verificationInProgress = false;
+  _whenCall = null;
+  _verifyCalls.clear();
+  _capturedArgs.clear();
+  _typedArgs.clear();
+  _typedNamedArgs.clear();
 }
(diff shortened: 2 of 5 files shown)


dskloetg:
Does the object on which we define behavior need to be the same as the object that implements the behavior? Maybe mocks can be created in pairs: one object on which we specify and verfiy the behavior from the test and one object we pass to the object under test. Then the object on which the behavior is defined doesn't need to implement the exact same interface as the mocked type but can accept matchers without causing errors.

srawlins:
OK this API is ready to merge, if we like it.

There is a lot of code, but it is actually nicely contained:

* the `typed` method (the public API),
* two private top-level objects, and a private top-level method,
* a top-level `resetMockitoState()` method, only used in testing,
* a new `FakeInvocation` class, where 95% of the work is done,
* one tiny check in `Mock.noSuchMethod`, to see if the `typed` API is invoked.

srawlins:
@dskloetg : I think using two different objects would be a large break from the mockito API. Something that Mockito gives us for free is method signatures. For example, the analyzer complains about the following:

```dart
class Cat {
  bool eatFood(List<String> foods) => true;
}

class MockCat extends Mock implements Cat {}

void fn() {
  var cat = new MockCat();
  when(cat.eatFood()).thenReturn(true);
}
```

Do you have a mockup of the API you had in mind?

README.md
```diff
@@ -147,6 +147,91 @@ expect(cat.sound(), "Purr");
 //using real object
 expect(cat.lives, 9);   
 ```
+
+## Strong mode compliance
+
+Unfortunately, the use of the arg matchers in mock method calls (like `cat.eatFood(any)`)
+violates the [Strong mode] type system. Specifically, if the method signature of a mocked
+method has a parameter with a parameterized type (like `List<int>`), then passing `any` or
```
dskloetg: Wouldn't this also happen if the parameter has a simple type?
"Unsound implicit cast from dynamic to String"

README.md
```diff
@@ -147,6 +147,91 @@ expect(cat.sound(), "Purr");
...
+Great! A little ugly, but it works. Here are some more examples:
+
+```dart
+when(cat.eatFood(typed/*<List<String>>*/(any), typed/*<List<String>>*/(any)))
+    .thenReturn(true);
+when(cat.eatFood(typed/*<List<String>>*/(argThat(contains("fish")))))
+    .thenReturn(true);
+```
+
+Named args require one more component: `typed` needs to know what named argument it is
```
dskloetg: If a name is passed to it, `typed` doesn't seem like the right name for this function as it's not the name of the type but rather the name of the parameter.
Maybe `arg` or `param` is more appropriate, or maybe argThat can be repurposed?

README.md
```diff
@@ -147,6 +147,91 @@ expect(cat.sound(), "Purr");
...
+being passed into:
+
+```dart
+when(cat.walk(
+    typed/*<List<String>>*/(any),
+    gaits: typed/*<Map<String, String>>*/(any), name: 'gaits')).thenReturn(true);
+```
+
+Note the `name` argument. Mockito should fail gracefully if you forget to name a `typed`
+call passed in as a named argument, or name the argument incorrectly.
```
dskloetg: So what should happen when you ommit `name: 'gaits'` on line 217?

lib/mockito.dart
```diff
@@ -48,6 +53,119 @@ class Mock {
   String toString() => _givenName != null ? _givenName : runtimeType.toString();
 }
 
+// Return a new [Invocation], reconstituted from [invocation], [_typedArgs],
+// and [_typedNamedArgs].
+Invocation _reconstituteInvocation(Invocation invocation) {
+  var newInvocation = new FakeInvocation(invocation);
+  _typedArgs.clear();
```
dskloetg: It looks like _typedArgs and _typedNamedArgs are being cleared without being used.
I realize now that FakeInvocation has access to them but it might be clearer to pass them into FakeInvocation to make it obvious how the invocation is reconstituted.

lib/mockito.dart
```diff
@@ -48,6 +53,119 @@ class Mock {
...
+  _typedNamedArgs.clear();
+  return newInvocation;
+}
+
+/// An Invocation class with a functional constructor.
```
dskloetg: What do you mean by "functional" constructor?

lib/mockito.dart
```diff
@@ -48,6 +53,119 @@ class Mock {
...
+  final Symbol memberName;
+  final Map<Symbol, dynamic> namedArguments;
+  final List<dynamic> positionalArguments;
+  final bool isGetter;
+  final bool isMethod;
+  final bool isSetter;
+
+  factory FakeInvocation(Invocation invocation) {
+    if (_typedArgs.isEmpty && _typedNamedArgs.isEmpty) {
+      return null;
```
dskloetg: Since this is only called when _typedArgs. or _typedNamedArgs is not empty, why not throw in this case?

lib/mockito.dart
```diff
@@ -48,6 +53,119 @@ class Mock {
...
+/// An Invocation class with a functional constructor.
+class FakeInvocation extends Invocation {
+  final Symbol memberName;
+  final Map<Symbol, dynamic> namedArguments;
+  final List<dynamic> positionalArguments;
+  final bool isGetter;
+  final bool isMethod;
+  final bool isSetter;
+
+  factory FakeInvocation(Invocation invocation) {
```
dskloetg: This constructor is really long. Maybe extract some functions `_reconstituteNameArgs` and `_reconstitutePositionalArgs`?

lib/mockito.dart
```diff
@@ -48,6 +53,119 @@ class Mock {
...
+Invocation _reconstituteInvocation(Invocation invocation) {
+  var newInvocation = new FakeInvocation(invocation);
+  _typedArgs.clear();
+  _typedNamedArgs.clear();
+  return newInvocation;
+}
+
+/// An Invocation class with a functional constructor.
+class FakeInvocation extends Invocation {
+  final Symbol memberName;
```
dskloetg: Add @override annotations.

dskloetg: Is it valid for a field to override a getter? I thought that wasn't allowed anymore. Or was it the other way around?

lib/mockito.dart
```diff
@@ -404,3 +531,14 @@ void logInvocations(List<Mock> mocks) {
     print(inv.toString());
   });
 }
+
+/// Should only be used during Mockito testing.
```
dskloetg: Could be good to express this in the name so it's easier to catch when someone uses it somewhere else.
`resetMockitoStateOnlyInMockitoTesting()`

README.md
srawlins: I don't think so. This is why it's not such a prominent problem. Just a little prominent:

```
$ cat lib/dogs.dart
import 'mockito.dart';

class Dog {
  bool eatFood(String food) => true;
  bool eatFood2(List<String> foods) => true;
}

class MockDog extends Mock implements Dog {}

void fn() {
  var dog = new MockDog();
  when(dog.eatFood(any)).thenReturn(true);
  when(dog.eatFood2(any)).thenReturn(true);
}
$ dartanalyzer --strong lib/dogs.dart
Analyzing [lib/dogs.dart]...
[warning] Unsound implicit cast from dynamic to List<String> (/Users/srawlins/code/dart-mockito/lib/dogs.dart, line 13, col 21)
1 warning found.
```

README.md
srawlins: I don't think I want to change `argThat`. I'm open to changing `name:` to something like `named:` or `asArg:` maybe?

README.md
srawlins: I think you're refering to the new test cases below:

* "should throw when [typed] used as a named arg, without `name:`"
* "should throw when [typed] used as a positional arg, with `name:`"
* "should throw when [typed] used as a named arg, with the wrong `name:`"

So that situation can be detected.

lib/mockito.dart
srawlins: Can you cite something regarding this? I wasn't sure either. The analyzer doesn't complain. I thought this was because the field adds an invisible, implicit getter. Do I need the `@override` annotations on such fields?

lib/mockito.dart
srawlins: Well... I don't know. I don't see any harm in making in public. It might be a feature for power users... or if someone writes improper mocks like my tests do... for some reason. Maybe other libraries that build on mockito's functionality.

lib/mockito.dart
srawlins: Done.

lib/mockito.dart
srawlins: Good point. Done!

lib/mockito.dart
srawlins: Good call. I've updated this.

README.md
dskloetg: I'd be curious to know why one is considered unsound but the other isn't.

README.md
dskloetg: How about changing `typed` to `argAs`?

lib/mockito.dart
dskloetg: Then maybe just remove the comment?

lib/mockito.dart
dskloetg: > Can you cite something regarding this?

I remember seeing an announcement or something about that but I can't find it anymore. I could only find https://github.com/dart-lang/sdk/issues/24928#issuecomment-157472898

> Do I need the `@override` annotations on such fields?

You never need `@override` annotations but it's good practice when you are overriding something. In this case these fields are overriding the getters on the extended class, so why not make that explicit?

lib/mockito.dart
```diff
@@ -48,6 +53,132 @@ class Mock {
   String toString() => _givenName != null ? _givenName : runtimeType.toString();
 }
 
+// Return a new [Invocation], reconstituted from [invocation], [_typedArgs],
+// and [_typedNamedArgs].
+Invocation _reconstituteInvocation(Invocation invocation) {
+  var newInvocation = new FakeInvocation(invocation);
```
dskloetg: just return without temp variable?

Andersmholmgren:
Just curious why an a new `typed` function is needed rather than making all the existing functions generic. 

Eg 

```
argThat/*<list<string>>*/(..)
```

etc

dskloetg:
I'm on vacation so I won't get back to this for 2.5 weeks. You can treat my remaining comments as FYI.

README.md
srawlins: I like the name `argAs`, except that this isn't _replacing_ `any` and `argThat`. It's wrapping them. I think this would look weird:

```dart
when(a.fn(argAs/*<List<String>>*/(argThat(equals('x')))));
when(a.fn(argAs/*<List<int>>*/(any)));
```

(not that any of this looks great)

lib/mockito.dart
srawlins: I think the comment is still valuable.

lib/mockito.dart
srawlins: Done.

srawlins:
@Andersmholmgren It's mostly just backwards-compatibility issues. If we changed `any` and `argThat` to return null instead of returning _ArgMatcher objects, then there would be a huge breaking change, where you could no longer use, e.g. `when(m.fn(any, null, any))`. I think there were more incompatibilities.

srawlins:
@TedSander This is ready to review/merge!

lib/mockito.dart
dskloetg: I don't understand. Either it's OK to be used outside or it's not.

README.md
dskloetg: What about it do you think looks weird?

dskloetg:
So we don't want to change `argThat` and `any` for compatibility, but wrapping is also a bit ugly.
What about introducing multiple new functions that replace `argThat` and `any` (and others) that do what their wrapped counterparts would do?

README.md
leafpetersen: Any implicit or explicit cast might fail in DDC.  Similarly, any implicit or explicit cast might fail in regular Dart checked mode.  In general, for simple types (like String), if the cast passed in checked mode, it will pass in DDC.  For complex types (like `List<String>`), DDC may throw a type error where checked mode wouldn't have.  This is because the checked mode check is unsound for composite types like `List<String>` (a `List<dynamic>` will be allowed through), and so DDC strengthens the cast.  Since these are places where code that worked in checked mode may fail in DDC, we emit a special  warnings for these casts that might fail in DDC where the would not have in checked mode.

README.md
```diff
@@ -147,6 +147,91 @@ expect(cat.sound(), "Purr");
...
+when(cat.eatFood(typed/*<List<String>>*/(argThat(contains("fish")))))
+    .thenReturn(true);
+```
+
+Named args require one more component: `typed` needs to know what named argument it is
+being passed into:
+
+```dart
+when(cat.walk(
+    typed/*<List<String>>*/(any),
```
leafpetersen: Not sure whether this approach is still under consideration, but I don't think you should need to pass these type arguments explicitly - they are in a position that should be inferrable.

TedSander: It is. Sam is just checking to make sure it will work with DDC. Do you see any reason it won't @leafpetersen

Removing those comments would be a nice improvement.

leafpetersen: If I'm understanding the overall approach here, I think this won't work without some compiler help.  The way that DDC generates code means that you can only get to noSuchMethod by going through dynamic.  So

```dart
dynamic cat = new MockCat();
cat.eatFood([]);
```

would work, but

```
var cat = new MockCat();
cat.eatFood([]);
}
```

won't, because we will infer that `cat` has type `MockCat`, and not generate the dynamic call.  It's arguably a bug that we still allow noSuchMethod to suppress the warnings that you should get by not fully implementing the `Cat` interface.  

I think this might be a place where some language and/or compiler help might be appropriate.

TedSander: Thanks Leaf exactly what we wanted to know. Some language or compiler help would be appreciated. Even if there are no changes to the language having recommendations from the DDC team on how we should solve this would be appreciated. I think Dart should have a solution to stubbing/mocking for tests.

leafpetersen: Filed a bug on the fact that we just silently accept this here:

https://github.com/dart-lang/sdk/issues/26863

Supporting this sort of thing is something I think we should do.  I'd like to be sure we understand and cover all (or as many as possible) of the use cases though.

README.md
srawlins: Where you declare 'arg' twice: `argAs(argThat)`

lib/mockito.dart
srawlins: It's OK to be used outside. Not by users of the mockito package though. Perhaps by developers extending the mockito package.

srawlins:
I can confirm that as of https://github.com/dart-lang/dev_compiler/commit/e5c2aac3fbb503fe0ede030dca9bfdc36bb0093a, the `typed` API in this pull request functions well with DDC. It's ready to roll.

dskloetg:
What do you think of my last suggestion from [22 days ago](https://github.com/fibulwinter/dart-mockito/pull/26#issuecomment-231717063)?

srawlins:
@dskloetg I'd have to make the following matching functions:

| Old | New |
|-----------|-----------------------------|
| **get** any | typedAny({String named}) |
| argThat(Matcher matcher) | typedArgThat(Matcher matcher, {String named}) |
| **get** captureAny | typedCaptureAny({String named}) |
| captureThat | typedCaptureThat(Matcher matcher, {String named}) |

This converts two getters to functions, only so they can be used with named args 😦 . I think this could have some real consequences, as passing `any` **calls** the `any` getter, but passing `typedAny` will now send the `typedAny` closure, **without** calling it. This mistake would also be hard (impossible?) to catch, as a closure could be a legitimate argument to a function.

I think the single `typed` function is a simpler API, and its only two characters longer (a `(` and a `)`).

lib/mockito.dart
```diff
@@ -48,6 +53,132 @@ class Mock {
...
+  return newInvocation;
+}
+
+/// An Invocation implementation that allows all attributes to be passed into
+/// the constructor.
+class FakeInvocation extends Invocation {
```
TedSander: I'm not really in love with this name, and the comment doesn't seem to really apply anymore. To a user they can't pass in all the attributes. Would TypedInvocation be better, or InvocationForTypedArguments? Fake doesn't really tell me much here.

lib/mockito.dart
```diff
@@ -48,6 +53,132 @@ class Mock {
...
+        namedArguments,
+        invocation.isGetter,
+        invocation.isMethod,
+        invocation.isSetter);
+  }
+
+  static Map<Symbol,dynamic> _reconstituteNamedArgs(Invocation invocation) {
+    var namedArguments = <Symbol, dynamic>{};
+    var _typedNamedArgSymbols = _typedNamedArgs.keys.map((name) => new Symbol(name));
+    invocation.namedArguments.forEach((name, arg) {
```
TedSander: A comment here would be helpful.

lib/mockito.dart
```diff
@@ -48,6 +53,132 @@ class Mock {
...
+      if (arg == null) {
+        if (!_typedNamedArgSymbols.contains(name)) {
+          // Incorrect usage of [typed], something like:
+          // `when(obj.fn(a: typed(any)))`.
+          throw new ArgumentError(
+              'A typed argument was passed in as a named argument named "$name", '
+              'but did not a value for its name. Each typed argument that is '
```
TedSander: but did not specify a value for its name.

lib/mockito.dart
```diff
@@ -48,6 +53,132 @@ class Mock {
...
+              'passed as a named argument needs to specify the `name` argument. '
+              'For example: `when(obj.fn(x: typed(any, name: "x")))`.');
+        }
+      } else {
+        // Add each real named argument that was _not_ passed with [typed].
+        namedArguments[name] = arg;
+      }
+    });
+
+    _typedNamedArgs.forEach((name, arg) {
```
TedSander: Comment please

lib/mockito.dart
```diff
@@ -48,6 +53,132 @@ class Mock {
...
+      Symbol nameSymbol = new Symbol(name);
+      if (!invocation.namedArguments.containsKey(nameSymbol)) {
+        // Incorrect usage of [name], something like:
+        // `when(obj.fn(typed(any, name: 'a')))`.
+        throw new ArgumentError(
+            'A typed argument was declared with name $name, but was not passed '
+            'as an argument named $name.');
```
TedSander: I think the example would be useful in the error.

lib/mockito.dart
```diff
@@ -48,6 +53,132 @@ class Mock {
...
+      }
+      if (invocation.namedArguments[nameSymbol] != null) {
+        // Incorrect usage of [name], something like:
+        // `when(obj.fn(a: typed(any, name: 'b'), b: "string"))`.
+        throw new ArgumentError(
+            'A typed argument was declared with name $name, but a different '
+            'value (${invocation.namedArguments[nameSymbol]}) was passed as '
+            '$name.');
```
TedSander: I think the examples you give in the comments are super helpful. Can we try to incorporate them into the error. I don't think long errors are a bad thing when they help developers.

lib/mockito.dart
```diff
@@ -48,6 +53,132 @@ class Mock {
...
+    var positionalArguments = <dynamic>[];
+    var nullPositionalArguments =
+        invocation.positionalArguments.where((arg) => arg == null);
+    if (_typedArgs.length != nullPositionalArguments.length) {
+      throw new ArgumentError(
+          'null arguments are not allowed alongside typed(); use '
+          '"typed(eq(null))"');
+    }
+    int i = 0;
+    int j = 0;
```
TedSander: A comment or name change for these would be helpful.
i => typedIndex
j => positionalIndex

test/mockito_test.dart
```diff
@@ -9,6 +9,14 @@ class RealClass {
   String methodWithNamedArgs(int x, {int y}) => "Real";
   String methodWithTwoNamedArgs(int x, {int y, int z}) => "Real";
   String methodWithObjArgs(RealClass x) => "Real";
+  // "SpecialArgs" here means type-parameterized args. But that makes for a long
+  // method name.
+  String methodWithSpecialArgs(
```
TedSander: I think I would rather describe "specialArgs" then call out it is a method in the name.

typeParameterizedFn maybe?

test/mockito_test.dart
```diff
@@ -9,6 +9,14 @@ class RealClass {
...
+      List<int> w, List<int> x, [List<int> y, List<int> z]) => "Real";
+  // "SpecialNamedArgs" here means type-parameterized, named args. But that
+  // makes for a long method name.
+  String methodWithSpecialNamedArgs(List<int> w, List<int> x, {List<int> y, List<int> z}) =>
```
TedSander: typeNameArgsFn maybe?

test/mockito_test.dart
```diff
@@ -204,6 +218,96 @@ void main() {
       when(mock.methodWithNormalArgs(argThat(equals(42)))).thenReturn("42");
       expect(mock.methodWithNormalArgs(43), equals("43"));
     });
+    test("should mock method with typed arg matchers", () {
+      when(mock.methodWithSpecialArgs(
+          typed/*<List<int>>*/(any), typed/*<List<int>>*/(any)))
```
TedSander: Can any of these tests or below remove the type comment like Leaf suggested in the README?

TedSander:
We should also update the minor version. Maybe while we are at it we should take it to 1.0.0

This code has been stable, and the dart team has asked us to reflect that fact.

lib/mockito.dart
srawlins: Done. _InvocationForTypedArguments.

lib/mockito.dart
srawlins: Done.

lib/mockito.dart
srawlins: Done.

lib/mockito.dart
srawlins: Done.

lib/mockito.dart
srawlins: Done.

lib/mockito.dart
srawlins: Done.

test/mockito_test.dart
srawlins: Done.

test/mockito_test.dart
srawlins: Done.

test/mockito_test.dart
srawlins: Done.

srawlins:
OK tests are good; new DDC tests in tool/.  I'm fine for bumping to 1.0.0. Should I go for it?

lib/src/mock.dart
```diff
@@ -55,6 +60,139 @@ class Mock {
   String toString() => _givenName != null ? _givenName : runtimeType.toString();
 }
 
+// Return a new [Invocation], reconstituted from [invocation], [_typedArgs],
+// and [_typedNamedArgs].
+Invocation _reconstituteInvocation(Invocation invocation) =>
```
TedSander: This seems like a little overkill, why not just inline the new?

lib/src/mock.dart
```diff
@@ -55,6 +60,139 @@ class Mock {
...
+
+  // Reconstitutes the named arguments in an invocation from [_typedNamedArgs].
+  //
+  // The namedArguments in [invocation] which are null should be represented
+  // by a stored value in [_typedNamedArgs]. The null presumably came from
+  // [typed].
+  static Map<Symbol,dynamic> _reconstituteNamedArgs(Invocation invocation) {
+    var namedArguments = <Symbol, dynamic>{};
+    var _typedNamedArgSymbols = _typedNamedArgs.keys.map((name) => new Symbol(name));
+    invocation.namedArguments.forEach((name, arg) {
```
TedSander: Comment for this forEach please.

tool/typed_mockito_test.dart
```diff
@@ -0,0 +1,497 @@
...
+///   bool fn(String) => true;
+/// }
+///
+/// when(obj.fn(any)).thenReturn(true);
+/// ```
+///
+/// In this code, `any` is certainly not a String, so it's not Strong-mode
+/// compatible, and DDC cannot work with it.
+///
+/// This file is a copy of `mockito_test.dart`, with all of the DDC-incompatible
```
TedSander: Why copy? Why not split the file?

tool/typed_mockito_test.dart
```diff
@@ -0,0 +1,497 @@
...
+    mock = new MockedClass();
+  });
+
+  tearDown(() {
+    // In some of the tests that expect an Error to be thrown, Mockito's
+    // global state can become invalid. Reset it.
+    resetMockitoState();
+  });
+
+  group("spy", () {
```
TedSander: Do we even want to test spy here? It shouldn't be used in DDC right?

tool/typed_mockito_test.dart
```diff
@@ -0,0 +1,497 @@
...
+      expect(mock.getter, equals("A"));
+    });
+    test("should mock hashCode", () {
+      named(mock, hashCode: 42);
+      expect(mock.hashCode, equals(42));
+    });
+    test("should have hashCode when it is not mocked", () {
+      expect(mock.hashCode, isNotNull);
+    });
+//    test("should n't mock toString", (){
```
TedSander: Remove?

TedSander:
Update the version number to 1.0.0??

tool/typed_mockito_test.dart
srawlins: The split would be kind of arbitrary. This subset is not just the typed API; it's everything that doesn't include `any`, `captureAny`, `argThat`, and `captureArg` outside of `typed`.

I want *all* the tests to pass using `pub run test`, and this subset to pass with DDC. There's also this weird case where setters are named differently in dart Invocations vs DDC Invocations (`==` vs `=`). So the mocked setter test is different:

```dart
// test/mockito_test.dart
mock.setter = "A";
expectFail("No matching calls. All calls: MockedClass.setter=A", () {
  verify(mock.setter = "B");
});
verify(mock.setter = "A");

// tool/typed_mockito_test.dart
mock.setter = "A";
expectFail("No matching calls. All calls: MockedClass.setter==A", () {
  verify(mock.setter = "B");
});
verify(mock.setter = "A");
```

tool/typed_mockito_test.dart
srawlins: why shouldn't it?

tool/typed_mockito_test.dart
srawlins: Phhh these files come with a lot of commented out code. Don't know what it's there for. Should I change it in `test/mockito_test.dart` and in `tool/typed_mockito_test.dart`?

tool/typed_mockito_test.dart
srawlins: Removed file.

tool/typed_mockito_test.dart
srawlins: Removed file.

lib/src/mock.dart
srawlins: Done.

lib/src/mock.dart
srawlins: Done.

srawlins:
Done. Version bump included.

TedSander:
Thanks Sam. Excited that we can now write DDC tests!

PR #28, opened 2016-08-02
Allow dart-mockito to be used by AOT (no dart:mirrors) clients

### LINKED ISSUE #27: Mockito without mirrors?
Would it be possible to create a version of mockito that doesn't use dart:mirrors? That would enable mockito to be used in environments that disable the mirrors support.

matanlurey:
- Moves reflection-based `spy` into `lib/src/spy.dart`
- Move the rest of the API to `lib/src/mock.dart`
- Removes use of `MirrorSystem` from `RealCall#toString`
- Update `.gitignore` to be a little more up to date

Closes #27

### DIFF
--- a/lib/src/spy.dart
+++ b/lib/src/spy.dart
@@ -1,0 +1,12 @@
+import 'dart:mirrors';
+
+import 'mock.dart' show CannedResponse, setDefaultResponse;
+
+dynamic spy(dynamic mock, dynamic spiedObject) {
+  var mirror = reflect(spiedObject);
+  setDefaultResponse(
+      mock,
+      () => new CannedResponse(null,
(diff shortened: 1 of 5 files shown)


matanlurey:
@TedSander :)

lib/src/spy.dart
```diff
@@ -0,0 +1,12 @@
+import 'dart:mirrors';
```
TedSander: Let's add a comment that this is separated so that we can keep mirrors separated.

TedSander:
Should we update the minor version?

matanlurey:
Done on both counts.

TedSander:
Thanks for the improvement Matan!

Issue #32, opened 2016-09-23
thenAnswer is not in the README.


srawlins:
It should be.

PR #33, opened 2016-09-23
Add docs for the argument matchers, and thenAnswer()

### LINKED ISSUE #30: please add method docs for typed, not just in the README
The README identifies a few situations in which you might want to use `typed`,
but nothing currently explains what `typed` actually is/does.
Please add method docs.

In particular, it seems to me that its purpose in life is to claim to strong mode that it returns whatever
type the call site demands (which of course can only happen if it always returns null... which it does).  And the reason we'd want that is basically to tell strong mode to shut up.
It's as if its name really should have been `inferTypeFromUsage`, although that would be much too long.

srawlins:
Fixes #30, #32

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,10 +23,7 @@ class Cat {
 }
 
 //Mock class
-class MockCat extends Mock implements Cat{
-  //this tells Dart analyzer you meant not to implement all methods, and not to hint/warn that methods are missing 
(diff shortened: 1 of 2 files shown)


TedSander:
Mind resolving the conflict and then merging?

srawlins:
Done!

PR #37, opened 2016-09-28
Add more types and documentation to Mock

matanlurey:
Mock is sort of a black-box of magic, and adding some documentation hopefully will help contributors in the future as well as give our users a better idea of what exactly they are using.

A tiny bit of refactoring on top of that, but no breaking API changes.

I've upped the pubspec to `+1`, but I can revert that if you'd prefer I don't.

/cc @srawlins


test/mockito_test.dart
```diff
@@ -33,13 +33,9 @@ abstract class AbstractFoo implements Foo {
   String baz();
 }
 
-class MockFoo extends AbstractFoo with Mock {
-  noSuchMethod(i) => super.noSuchMethod(i);
```
srawlins: omg thank you; been meaning to do this forever.

srawlins:
LGTM; CC @TedSander

test/mockito_test.dart
matanlurey: 🍔

TedSander:
LGTM

PR #39, opened 2016-09-29
Tighten public API + cleanup 

thso:
Explicitly expose public methods in order not to leak internals and re-expose `mockito_no_mirrors.dart` from `mockito.dart` to ensure consistency.

Removes `CannedResponse`,  `InOrderVerification`, `Verification`, `RealCall`, `resetMockitoState` from the public API since these are only meant to be used internally by Mockito.

+ simplify `toString` logic in `RealCall`,
+ have `verifyInOrder` take a `List` since taking a single argument would be an invalid usage,
+ other minor adjustments and `dartfmt`

### DIFF
--- a/lib/src/mock.dart
+++ b/lib/src/mock.dart
@@ -526,15 +526,15 @@ class VerificationResult {
 
 typedef dynamic Answering(Invocation realInvocation);
 
-typedef VerificationResult _Verification(matchingInvocations);
+typedef VerificationResult Verification(matchingInvocations);
 
 typedef void _InOrderVerification(List<dynamic> recordedInvocations);
 
-_Verification get verifyNever => _makeVerify(true);
+Verification get verifyNever => _makeVerify(true);
 
-_Verification get verify => _makeVerify(false);
+Verification get verify => _makeVerify(false);
 
-_Verification _makeVerify(bool never) {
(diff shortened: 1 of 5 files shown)


thso:
Removing these from the public API is a breaking change, so we should bump the major version. I checked and there are no references at all to these internally in Google. Going forward we should be more careful with what's exposed.

WDYT @TedSander ?

TedSander:
I think this looks good, but yes please bump the major version as it does have breaking changes.

TedSander:
Please rebase. Matan just added matcher functionality that clients can use.

matanlurey:
+1. Thanks.

lib/src/mock.dart
```diff
@@ -528,15 +526,15 @@ class VerificationResult {
 
 typedef dynamic Answering(Invocation realInvocation);
 
-typedef VerificationResult Verification(matchingInvocations);
+typedef VerificationResult _Verification(matchingInvocations);
```
srawlins: I think this has to stay public, yes?

Verification verification = verify(foo.bar(captureAny));
var args = verification.captured;
verification.called(1);

thso: Done, thanks.

thso:
PTAL @srawlins

PR #42, opened 2016-10-07
Move all mirrors-based API to mirrors.dart

matanlurey:
Updated `CHANGELOG.md`, and removed need to import `mockito_no_mirrors.dart` for Flutter.

# Breaking Changes

* To use `spy()`, you must import `package:mockito/mirrors.dart`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 ## 2.0.0-dev
 
-* Deprecated usage of `spy` and any `dart:mirrors` based API. Users may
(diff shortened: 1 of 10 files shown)


srawlins:
Why are we fully deprecating it? Mirrors aren't going anywhere.

TedSander:
Agreed. Talked to Matan on a separate thread. He will update with that feedback.

matanlurey:
Updated.

TedSander:
Thanks for the update.

PR #46, opened 2016-12-19
Cleanup

kevmoo:
How close are we to releasing 2.0.0?

I'd like to get the GitHub URL fixed on the pub site, etc

### DIFF
--- a/.analysis_options
+++ b/.analysis_options
@@ -1,0 +1,29 @@
+analyzer:
+ strong-mode: true
+linter:
+  rules:
+     # Errors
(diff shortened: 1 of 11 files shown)


.analysis_options
```diff
@@ -0,0 +1,29 @@
+analyzer:
+ strong-mode: true
+linter:
+  rules:
```
srawlins: Where does this list come from? Is this a published list, like from Effective Dart or something?

srawlins:
LGTM

PR #48, opened 2016-12-28
Add new throwOnMissingStub method

### LINKED ISSUE #47: Friendly debug messages when mock methods without expectations set are called.
For example:
```
//Real class
class Cat {
  String sound() => "Meow";
}

//Mock class
class MockCat extends Mock implements Cat {}

//mock creation
var mockCat = new MockCat();

// Would throw exception "NoSuchMethodError: The method 'toUpperCase' was called on null.".
mockCat.sound().toUpperCase();
```

The actual cause is during the test development, developer forgot to write "when(cat.sound()).thenReturn("Purr");".

The message "'toUpperCase' was called on null" isn't helpful for finding out the casue and a more friendly message could be "'toUpperCase' was called on the return value of 'Cat.sound()', but no exceptions were set on it".

This could be achieved by returning instead of 'null' but a special object carrying the method invocation information for any 'un-expected' methods.

I can come with a design doc if this sounds possible.

srawlins:
Fixes #47 

Mostly. It'd be nice to have a global one too, but this was simple to implement for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.0.1
+
+* Add a new `throwOnMissingStub` method to the API.
+
 ## 2.0.0
 
 * Removed `mockito_no_mirrors.dart`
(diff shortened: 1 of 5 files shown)


kevmoo:
@srawlins Take a look at Travis?

srawlins:
Travis green

PR #49, opened 2017-01-03
Cleanup for v2.0.0

matanlurey:
Major change is removing `mockito_no_mirrors.dart`


srawlins:
o_O So Mockito is back to a mirrors-only library?

matanlurey:
@srawlins Mirrors is only via `package:mockito/mirrors.dart`.

PR #58, opened 2017-03-17
Add types to any and captureAny getters

Anrock:
Hi. Just noticed, that ```when(foo.bar(typed(any)))``` produces
```[error] The argument type 'dynamic' can't be assigned to the parameter type 'ArgMatcher'.```
Dunno why, but ```get any => new ArgMatcher(anything, false);``` inferred type is dynamic.
So i just added ```ArgMatcher``` type to these getters.

### DIFF
--- a/lib/src/mock.dart
+++ b/lib/src/mock.dart
@@ -518,11 +518,11 @@ class ArgMatcher {
 }
 
 /// An argument matcher that matches any argument passed in "this" position.
-get any => new ArgMatcher(anything, false);
+ArgMatcher get any => new ArgMatcher(anything, false);
 
 /// An argument matcher that matches any argument passed in "this" position, and
 /// captures the argument for later access with `captured`.
-get captureAny => new ArgMatcher(anything, true);
+ArgMatcher get captureAny => new ArgMatcher(anything, true);
 
 /// An argument matcher that matches an argument that matches [matcher].
 argThat(Matcher matcher) => new ArgMatcher(matcher, false);

### CODE REVIEW

**Comment by Anrock** [comment]
Just signed CLA.

**Comment by Anrock** [comment]
Well, probably it's not that simple - looks like it works only for dart 1.22. Any advice?

**Comment by TedSander** [comment]
@srawlins have any ideas on this? We have it working internally. I'm not sure what is going on here. @Anrock what version are you using? and with package:test or package:unittest?

**Comment by Anrock** [comment]
@TedSander sorry for delay. Currently i'm using Dart 1.22.1, mockito 2.0.2 with test package of 0.12.20+3 (it was mockito 1.0.1 and test 0.12.15+2 when i created pull request), it's still reproduces.
You have to add
```
analyzer:
  strong-mode:
    implicit-dynamic: true
```
to .analysis_options for it to produce error.

TedSander:
I'm not exactly sure what is going on here. We have tests that are working with this. Internally we just use the analysis_options of:
analyzer:
  strong-mode: true

but when I passed the no-implicit-dynamic to dartanalyzer it doesn't give an error on the file. When I drop the flag it doesn't give an error either. We use dart 1.23 internally at the moment.

Could it be that the typed function is coming from a different library other than mockito? I would expect the type inference to work with this setup.

@jmesserly Jennifer any ideas?

Anrock:
Well, that strange. I'll try to make a minimal repro project tommorow, then.

jmesserly:
> [error] The argument type 'dynamic' can't be assigned to the parameter type 'ArgMatcher'.
> Jennifer any ideas?

I think this error comes from strong mode and --no-implicit-casts. It makes sense, yes. The fix would be to add a type on the getter (like this pull request does) or add a cast: `typed(any as ArgMatcher)`

srawlins:
I think to accept this, we'd have to split the analysis, and tests, into two parts. Or go full on strong mode.

Basically much of the tests do not pass new strong mode, so we'd have to start adding `typed` everywhere. Or make a weak mode test, showing examples of mockito tests still working in weak mode, and a strong mode test, with tests that can be run in DDC, and can be analyzed by strong mode analyzer.

jmesserly:
@srawlins can you clarify?

> Basically much of the tests do not pass new strong mode

I think this error only happens when someone opts in to --no-implicit-casts analysis flag. At least that's my understanding, I could be missing something here.

srawlins:
Sorry for leaving this.

@jmesserly By "much of the tests do no pass new strong mode", I just meant that we don't analyze the test files with strong, because most of the tests would fail.

See the [Travis failure](https://travis-ci.org/dart-lang/mockito/jobs/212067479)? When we leave the type of `any` as dynamic (or... whatever... no type), then all of the weak-mode-compliant tests look fine (the code is still actually wonky, because yes, `any` _does_ return an ArgMatcher). But if we declare that `any` returns an ArgMatcher, then the analyzer pitches a fit about every unwrapped-with-`typed` call to any.

So we do need to accept this pull request, but cannot until mockito is Dart 2.0/Strong-only. I'll merge this after #71 is fixed (I plan to do soon).

matanlurey:
Assumed stale.

PR #59, opened 2017-04-01
Adding a "mockito + package:test" example.

Sfshaza:
@kevmoo, what is required to run testing on this test?

Please review, @kevmoo, @isoos!

### DIFF
--- a/test/example/iss/README.md
+++ b/test/example/iss/README.md
@@ -50,10 +50,10 @@ between them.
 ```
 
 **ISS Spotter**
-: Stubs `IssLocator.currentPosition()` using `when().thenReturn()`.
-Calling currentPosition() then returns the predefined location for the
-space station. Evaluate whether the space station is visible from a
-second predefined location. This test runs asynchronously.
+: Stubs `IssLocator.currentPosition` using `when().thenReturn()`.
+Evaluate whether the space station (using a predefined location)
+is visible from a second predefined location.
+This test runs asynchronously.
 
 ```
   group('ISS spotter', () {
@@ -61,6 +61,7 @@ second predefined location. This test runs asynchronously.
       Point<double> sf = new Point(37.783333, -122.416667);
       Point<double> mtv = new Point(37.389444, -122.081944);
       IssLocator locator = new MockIssLocator();
+      // Mountain View should be visible from San Francisco.
       when(locator.currentPosition).thenReturn(sf);
 
       var spotter = new IssSpotter(locator, mtv);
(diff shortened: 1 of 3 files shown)


kevmoo:
> what is required to run testing on this test?

Looks like they already did!

test/example/iss/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+**IssSpotter**
+: Performs calculations from the observer's location on earth.
+
+---
+
+The unit test, `iss_dart.test`, mocks the IssLocator class:
+
+```
+@proxy
```
srawlins: Should this be explained here? I don't know what this is.

test/example/iss/README.md
```diff
@@ -0,0 +1,81 @@
...
+      Point<double> sf = new Point(37.783333, -122.416667);
+      Point<double> mtv = new Point(37.389444, -122.081944);
+      double d = sphericalDistanceKm(sf, mtv);
+      expect(d, closeTo(52.8, 0.1));
+    });
+  });
+```
+
+** ISS Spotter**
+: Stubs `IssLocator.currentPosition()` using `when().thenReturn()`.
```
srawlins: I don't think currentPosition should have parens here. Would be confusing when below we see it without parens, as if it were a reference to a method, rather than a getter call.

test/example/iss/iss_test.dart
```diff
@@ -0,0 +1,70 @@
...
+  // second predefined location. This test runs asynchronously.
+  group('ISS spotter', () {
+    test('ISS visible', () async {
+      Point<double> sf = new Point(37.783333, -122.416667);
+      Point<double> mtv = new Point(37.389444, -122.081944);
+      IssLocator locator = new MockIssLocator();
+      when(locator.currentPosition).thenReturn(sf);
+
+      var spotter = new IssSpotter(locator, mtv);
+      expect(spotter.isVisible, true);
```
srawlins: nit: I would prefer a comment just above this line stating something like:

> Mountain View should be visible from San Francisco.

test/example/iss/iss_test.dart
```diff
@@ -17,7 +17,7 @@ import 'dart:math';
 import 'package:mockito/mockito.dart';
 import 'package:test/test.dart';
 
-import 'package:testing_example/iss.dart';
+import 'iss.dart';
```
kevmoo: Woo hoo!

Make sure you squash before you commit...

Sfshaza:
Thanks for your feedback, @srawlins!

PR #69, opened 2017-07-31
Address the verify(...).called(0) issue. Also:

srawlins:
* Document verifyNever
* Improve the error message of verifications
* Add some verification tests


srawlins:
CC @jonahwilliams

PR #72, opened 2017-08-09
Add ability to wait for an invocation

jdicker:
Adding a way to wait for an invocation. This is particularly useful when testing event based functionality where there is no easy way to await the expected behavior.

### DIFF
--- a/test/mockito_test.dart
+++ b/test/mockito_test.dart
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import 'dart:async';
+
 import 'package:mockito/mockito.dart';
 import 'package:mockito/src/mock.dart'
     show resetMockitoState, throwOnMissingStub;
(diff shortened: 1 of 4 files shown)


jdicker:
I signed the CLA.

Great name suggestions, they definitely read a lot nicer than 'await waitFor'. I'll change it to untilCalled, I think that one speaks to me the most.

srawlins:
Looks good, are the changes complete?

srawlins:
I think you need to dartfmt. Feel free to run the travis check locally, tool/travis.sh.

jdicker:
Yeah, I didn't really do a very good job of setting up a standard environment so I've been trying to just use whatever the UI exposes here which is working...not great. I'll set up a regular environment and do the checks myself.

test/mockito_test.dart
```diff
@@ -352,6 +381,210 @@ void main() {
...
+    });
+
+    group("on methods not yet called", () {
+      setUp(() {
+        new RealClassController(mock, streamController);
+      });
+
+      test("waits for method without args", () async {
+        streamController.add(new CallMethodsEvent());
+        verifyNever(mock.methodWithoutArgs());
```
srawlins: This is kind of tricky behavior, but I think it is valid today. This relies on the fact that `streamController.add(new CallMethodsEvent())` will call `mock.MethodWithoutArgs()` **in the next microtask** (or later, whatever). However, I think it's not actually specified whether this is the case, at the language level. E.g. ([dartpad](https://dartpad.dartlang.org/f5a0d0a940fe5778fbf6b9f04cf88fe9)

```dart
import 'dart:async';

main() async {
  var a = 1;
  new Future.value(null).then((_) => a = 2);
  print(a);
  await null;
  print(a);
}
```

prints `1, 2` today, but is not guaranteed to. All implementations today, however, _do_ execute that `a = 2` in the next microtask, so we're safe today. If Dart someday breaks this, it will be limited to these tests, not the implementation.

jdicker: Agreed, this test is brittle in the sense of it might break in the future with changes to dart even if the underlying behavior of untilCalled is still correct. I think we're on the same page, but just to expand a bit on my thinking here - IMO it's better to ensure that I'm actually testing the behavior than make the test less prone to future breakages due to dart updates. If I removed the verifyNever this test would be prone to issues where if untilCalled(...) didn't actually wait until the method was called and instead returned an immediate future, the test might still pass if the method happened to have been called earlier.

srawlins: Agreed. I think the test is good enough (and important). Just made me scratch my head. :smile:

jdicker:
Added a section to the README, but open to suggestions if I didn't give examples that make things clear enough.

srawlins:
Thanks!

PR #81, opened 2017-10-24
cleanup

a14n:


lib/src/mock.dart
```diff
@@ -726,7 +726,7 @@ typedef PostExpectation Expectation(x);
 /// `when` or `verify` context (a call "for real"), Mockito will respond with
 /// the stored canned response, if it can match the mock method parameters.
 ///
-/// The response generators include [thenReturn], [thenAnswer], and [thenThrow].
+/// The response generators include `thenReturn`, `thenAnswer`, and `thenThrow`.
```
srawlins: The `[]` is how dartdoc links to other methods... why is this changed?

a14n: There's a lint or hint :

> [dart] Only reference in scope identifiers in doc comments.

Actually all those methods are on `PostExpectation`. Would you prefer `[PostExpectation.thenThrow]` ?

srawlins: Boo, we need [this feature](https://github.com/dart-lang/markdown/issues/172) in markdown, in order for it to do what I'd want. Ok you're change is good then. Thanks!

PR #85, opened 2017-11-16
Flip to Dart 2.0-only syntax

srawlins:
TODO:

* [x] test internally
* [x] write `upgrading-to-3.0.0.md`

Summary:

* The old `typed` system of storing argument matchers is now the only system 😄  `typed` goes away.
* README is updated.
* Breaking changes:
  * `any`, and `captureAny` cannot be used for named arguments; instead `anyNamed()` and `captureAnyNamed()` must be used ☹️ .
  * An argument matcher passed as a named argument must declare it's name. `x: argThat(isNotNull)` must be `x: argThat(isNotNull, named: 'x')`. ☹️ 
  * `null` cannot be passed as a named argument.
  * `null` cannot be passed as a positional argument when argument matchers are used in the same invocation.
* Update tests. Add a few new tests.
* `typed` is now a no-op: it returns it's first argument.

The upgrade path is not super obvious, so I want to write a upgrading-to-3.0.0.md document. Things like `x: typed(any, named: 'x')` becomes `x: anyNamed('x')`, etc.


srawlins:
Hi @matanlurey and @TedSander . No rush, but this is essentially ready for review. The code and tests are complete. I'm looking over the README though and want to write an upgrade path doc.

I'm also thinking I want to hold off on releasing for a few weeks/months while the void type stuff makes it into Dart dev releases, and maybe can make use of the planned new Invocation constructor and delete the `_InvocationForMatchedArguments` class.

README.md
```diff
@@ -124,6 +129,36 @@ If an argument other than an ArgMatcher (like `any`, `anyNamed()`, `argThat`,
...
+verify(cat.hunt(argThat(contains("yard")), null)); // BAD: adjascent null.
+verify(cat.hunt(argThat(contains("yard")), argThat(isNull))); // OK: wrapped in ArgMatcher.
+verify(cat.eatFood("Milk", hungry: null)); // BAD: null as named argument.
+verify(cat.eatFood("Milk", hungry: argThat(isNull))); // BAD: null as named argument.
+```
+
+## Named arguments
+
+Mockito currently has an awkward nuisance to it's syntax: named arguments and
+argument matchers require more specification than you might think. Namely, you
```
TedSander: I'm wondering if we can do the 1000 foot explaination on this. Something like: This is because the language doesn't provide the name of the argument in the invocation and we can't rely on the position of a named argument.

srawlins: Good call. Done.

srawlins:
Rebased and tested on Dart 2.0.0-dev.19.0. Looks good.

chalin:
This PR is happily approved by all reviewers. Might it be merged soon?

I've hit an issue with Angular 5 example app component tests under Dart 2-dev:

```nocode
Type 'ArgMatcher' is not a subtype of type 'String'
```

because of

```dart
    var c = verify(mockRouter.navigate(captureAny, captureAny));
```

cc @kwalrath

srawlins:
I need to check off those two boxes at the top, but I hope to focus next week on Mockito and get 3.0.0 out the door.

For your issue above with DDC, you need the pre-3.0 typed API:

```dart
var c = verify(mockRouter.navigate(typed(captureAny), typed(captureAny)));
```

srawlins:
OK this is looking good. I'm going to do a spot check internally, and then I'll land this to master.

PR #88, opened 2017-11-29
Throw an ArgumentError if `thenReturn` is called with futures or streams.

### LINKED ISSUE #79: Push users to .thenAnswer(<Future|Stream>)?
This is what I see a lot:

```dart
var mockThing = new MockThing();
when(mock.isDone).thenReturn(new Future.value(true));

// ...
runTest(thing: mockThing);
```

But it can break in basically un-debuggable ways, especially around zones. Normally a `Future` or `Stream` won't be created _until_ an invocation is made, but with the `.thenReturn` pattern, it has been created in the test setup.

The "correct" behavior here would be to use `.thenAnswer`:

```dart
var mockThing = new MockThing();
when(mock.isDone).thenAnswer((_) => new Future.value(true));
```

I don't know if this a documentation, API, or just UX issue, or there is even a _real_ issue - maybe mockito alone is mysterious so nobody assumes code to work great 100% of the time. We could make the API more verbose, i.e.:

```dart
thenAnswerSync<T>(T value);
thenAnswerAsync<T>(T Function() run);
```

But I don't know if that would help unless we threw on an invalid type at runtime.

alanrussian:
Also, documented this behavior in the README.

This fixes #79.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 2.3.0
 
-* `thenReturn` now asserts that neither a `Future` nor `Stream` is provided.
-  `thenReturn` calls with futures and streams should be changed to
+* `thenReturn` now throws an `ArgumentError` if either a `Future` or `Stream`
+  is provided. `thenReturn` calls with futures and streams should be changed to
   `thenAnswer`. See the README for more information.
 
 ## 2.2.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.3.0
+
+* `thenReturn` now asserts that neither a `Future` nor `Stream` is provided.
+  `thenReturn` calls with futures and streams should be changed to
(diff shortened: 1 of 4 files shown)


alanrussian:
I updated the change log to document this since it's a breaking change. I just put down version number 2.3.0. Let me know if I should change it.

alanrussian:
Looks like this is failing on Dart 1.21.1, because [assert message support wasn't added until Dart 1.22](https://news.dartlang.org/2017/02/dart-122-faster-tools-assert-messages.html).

I think providing a message is valuable, so I'll change this to return an ArgumentError.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.3.0
```
matanlurey: Sort of late, but this should be 3.0.0.

TedSander: Fix here... the pubspec wasn't updated either: https://github.com/dart-lang/mockito/pull/89

PR #90, opened 2018-01-29
Capture T in `verify`+ so T may be `void` in dart 2: `verify(m.void())`

MichaelRFairhurst:
While the semantics here are not yet fully decided, it is likely that
dart 2 will disallow passing void into dynamic. The most backwards
compatible solution here is to parameterize meta function types over T
so that the type `void` is inferred, *possibly* allowing the result of a
void method to be passed in.

This change unblocks us for incrementally rolling out the new void
semantics.

It may in fact require a change to specify the `void` type directly,
rather than allowing `T` to be void for any `T`, but for now that's not
required and would hurt backwards compatibility since `void` didn't used
to be allowed there.

It also may not be allowed at all under any case, in which case we'd be
in for a fun ride.

### DIFF
--- a/lib/src/mock.dart
+++ b/lib/src/mock.dart
@@ -607,12 +607,8 @@ class VerificationResult {
 
 typedef dynamic Answering(Invocation realInvocation);
 
-/// Future-proofing: `<T>` may, in dart 2, allow assignment of "void", so that
-/// you can do `verify(m.voidFn())`. Subject to change.
 typedef Verification = VerificationResult Function<T>(T matchingInvocations);
 
-/// Future-proofing: `<T>` may, in dart 2, allow assignment of "void", so that
-/// you can do `verifyInOrder([m.voidFn(), m.voidFnToo()])`. Subject to change.
 typedef _InOrderVerification = void Function<T>(List<T> recordedInvocations);
 
 /// Verify that a method on a mock object was never called with the given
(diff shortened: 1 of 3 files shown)


lib/src/mock.dart
```diff
@@ -607,9 +607,13 @@ class VerificationResult {
 
 typedef dynamic Answering(Invocation realInvocation);
 
-typedef VerificationResult Verification(matchingInvocations);
+/// Future-proofing: `<T>` may, in dart 2, allow assignment of "void", so that
```
srawlins: You can remove these comments or at least make them non-doc (`//`).

matanlurey:
> Can you also bump the minimum SDK version to 1.24.0, and remove 1.21.1 from .travis.yml, as per dart-lang/sdk:CHANGELOG.md@master#1240---12-06-2017, this is when the new Function syntax was introduced.

This isn't correct, you need `2.0.0-dev`, the analyzer crashes on 1.24.0.

matanlurey:
I'm honestly not sure

srawlins:
Alright, in the spirit of I-dont-want-to-think-about-this, let's bump the minimum SDK version to 2.0.0-dev.16.0. It should be fine:

* It's pretty recent; I'm pretty sure new Function syntax is happy in analyzer -dev.16.0,
* It's the version Flutter is currently running on; solid release,
* We're not going to release with these fixes for a while, first will come a pre-release; this PR mostly needs to get in soon for an internal bump.

MichaelRFairhurst:
not to complicate your firm decision, but I think the problems with the new
function types were limited to specific cases (not here), such as using
them as a type argument of a const value, or as a type parameter bound, etc.

But yeah, I wanted to do some research of when this works, and its probably
easier just to call it at 2.0.0-dev.16.0, if there are no other problems
with doing so.

>

MichaelRFairhurst:
changes pushed up, thanks for the review guys!

srawlins:
Thanks, Mike, looks good!

PR #92, opened 2018-02-08
verify*Interactions methods throw helpfully that they expect Mock

srawlins:
So, the alternative is to Use-The-Damn-Language-Like-You're-S'posed-To-™ and type the incoming argument.

However, when playing around with this internally, I think it actually gets in the way a lot. You see this a lot:

```dart
main() {
  HttpClient httpClient;

  setUp(() {
    httpClient = new MockHttpClient();
  });

  test('foo', () {
    // ...
    verifyZeroInteractions(httpClient);
  });
}

class MockHttpClient extends Mock implements HttpClient {}
```

Which would start throwing static analysis errors, because `httpClient` is declared as a `HttpClient`, not a `Mock`. You'd have to change the type at the declaration to `MockHttpClient`.

(A) This would be a lot of work to upgrade existing code, for little benefit. (B) I think this produces more overhead for a developer than is helpful: you want to think of your mock objects as really almost basically just the same thing as the type they are mocking. Typing all your mock objects to the type that they implement is intuitive (it's done everywhere).

### DIFF
--- a/lib/src/mock.dart
+++ b/lib/src/mock.dart
@@ -710,18 +710,24 @@ void _throwMockArgumentError(method) =>
     throw new ArgumentError('$method must only be given a Mock object');
 
 void verifyNoMoreInteractions(var mock) {
-  if (mock is! Mock) _throwMockArgumentError('verifyNoMoreInteractions');
-  var unverified = mock._realCalls.where((inv) => !inv.verified).toList();
-  if (unverified.isNotEmpty) {
-    fail("No more calls expected, but following found: " + unverified.join());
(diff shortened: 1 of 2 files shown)


matanlurey:
Though, you probably want to write the code like this:

```dart
if (mock is Mock) {
  // use
} else { 
 // throw
}
```

... to take advantage of type promotion.

srawlins:
Lol I sort of did it this way in silent protest that this way doesn't already feature type promotion! https://github.com/dart-lang/sdk/issues/25565

I can change it though.

PR #94, opened 2018-02-09
Bump to 2.2.2

srawlins:
I've made a branch (naming is hard; it's called `two-two-one-plus`) for a quick 2.2.2 release for Flutter, for https://github.com/flutter/flutter/issues/14532.

@a-siva thinks we just need the USES_DYNAMIC_AS_BOTTOM fix, the only commit in this branch after the 2.2.1 bump.

(And I hope this is the "correct" way to patch "old" releases like this...)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 2.2.2
+
+* Remove last violation of `USES_DYNAMIC_AS_BOTTOM` error. This should make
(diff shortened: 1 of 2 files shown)


srawlins:
(Ah, and I realize now I should have pushed this branch before cherry-picking the uses-dynamic-as-bottom fix. Alas.)

PR #95, opened 2018-02-10
Bump 2.2 branch to 2.2.3, fixing compliance with SDK 2.0.0-dev.23.0

srawlins:
CC @MichaelRFairhurst

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -6,7 +6,7 @@ authors:
 description: A mock framework inspired by Mockito.
 homepage: https://github.com/dart-lang/mockito
 environment:
-  sdk: '>=1.21.0 <2.0.0'
+  sdk: '>=2.0.0-dev.16.0 <2.0.0'
 dependencies:
   collection: '^1.1.0'
   matcher: '^0.12.0'
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: mockito
-version: 2.2.1
+version: 2.2.2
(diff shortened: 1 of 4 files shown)


pubspec.yaml
```diff
@@ -1,12 +1,12 @@
...
-version: 2.2.2
+version: 2.2.3
 authors:
   - Dmitriy Fibulwinter <fibulwinter@gmail.com>
   - Dart Team <misc@dartlang.org>
 description: A mock framework inspired by Mockito.
 homepage: https://github.com/dart-lang/mockito
 environment:
-  sdk: '>=1.21.0 <2.0.0'
+  sdk: '>=2.0.0-dev.16.0 <2.0.0'
```
MichaelRFairhurst: This is further backwards compatible than this...supposedly as far back as 1.24.0. However, its possible there were still bugs at that time. I'll do some experiments, otherwise I think the safest restriction is dev.16

srawlins: OK I'll leave it at 1.24.0; no features, just a compatibility release.

chalin:
Will this be ported to 3.0.0-alpha too? /cc @kwalrath

srawlins:
I'm pretty sure this is all in 3.0.0-dev.

chalin:
When I switch to 3.0.0-alpha I get the "void" error that we used to get before this fix.

srawlins:
Yeah not -alpha. Just -dev. Unreleased.

PR #96, opened 2018-02-13
First draft of upgrade guide for Mockito 3.0

srawlins:


upgrading-to-mockito-3.md
```diff
@@ -0,0 +1,142 @@
...
+
+```dart
+when(cat.eatFood(argThat(contains('mouse')), hungry: any))...
+```
+
+In this code, Mockito is not passing a String as the first argument to
+`eatFood()`, nor a bool as `hungry:`, which is what the method expects.
+Instead, Mockito passes an ArgumentMatcher, a completely different object from
+a String or bool! This is no longer allowed in Dart 2.  As an interim solution,
+Mockito 1.0.0 sports an awkward `typed` function, which wraps other Mockito
```
matanlurey: Is this true, was it just for Mockito 2.x, right?

srawlins: Mockito 1 is when we introduced `typed`. Mockito 2 is when you un-exported `spy`.

PR #100, opened 2018-02-16
Try using a staged, fancy travis config.

matanlurey:
We can keep this, or not. I'm assuming we want to test this package with DDC, so.


matanlurey:
Completing this now

matanlurey:
@srawlins PTAL again.

Notable changes:
* I fixed _every_ test to work with DDC, I realize some tests are duplicates now.
* The `<T>` return type for `thenAnswer` etc was not strong-mode clean; it tried to return the `MockFoo` object where `T` was expected (i.e. `String`). I fixed it.

.travis.yml
```diff
@@ -1,5 +1,47 @@
 language: dart
-sudo: false
-dart:
-  - dev
-script: ./tool/travis.sh
+
+# Gives more resources on Travis (8GB Ram, 2 CPUs).
+# Do not remove without verifying w/ Travis.
+sudo: required
```
kevmoo: We should update `mono_repo` to handle "normal" repos – just to support generating this config.

I chatted w/ @natebosch about this yesterday
CC @alorenzen

matanlurey:
I can confirm this works fine locally.

I think this is the same issues that @grouma @nex3 @alorenzen are seeing for AngularDart (timeouts for the test runner). So whatever is the fix there should also be the fix here.

nex3:
We were seeing this with Chrome, not Firefox. If it's happening with both browsers, that's very interesting information.

matanlurey:
@nex3 Different error for Firefox:

```
00:00 +0 -1: compiling test/mockito_test.dart [E]                              
  Failed to load "test/mockito_test.dart":
  Failed to run Firefox: Firefox failed with exit code 1.
  Standard output:
  Error: GDK_BACKEND does not match available displays
  .
  package:test/src/runner/browser/browser.dart 108           new Browser.<fn>
```

matanlurey:
@srawlins FYI, this should work now. I am going to merge on green.

PR #101, opened 2018-02-16
Generic support for `thenReturn` and `thenAnswer`

grouma:
I have been running into these errors as I have migrated folks over to DDC. Thankfully they can be caught through analysis instead of runtime. Unfortunately, it will be a huge pain in the ass to pull into Google3. We'll see who that works lands on :P

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: mockito
-version: 3.0.0-alpha
+version: 3.0.0
 authors:
   - Dmitriy Fibulwinter <fibulwinter@gmail.com>
   - Dart Team <misc@dartlang.org>
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: mockito
-version: 3.0.0
+version: 3.0.0-dev
 authors:
   - Dmitriy Fibulwinter <fibulwinter@gmail.com>
   - Dart Team <misc@dartlang.org>
--- a/test/mockito_test.dart
+++ b/test/mockito_test.dart
@@ -260,8 +260,8 @@ void main() {
     });
 
     test("should return mock to make simple oneline mocks", () {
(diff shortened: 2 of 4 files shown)


TedSander:
I'd rather not land the change until we figure out who is going to do the work. We've had problems with packages getting way out of sync because no-one signed up for landing the patch.

grouma:
That's fair. That being said landing this change sooner than later will be beneficial for the Dart 2.0 work. These issues will crop up once the VM has the new semantics.

TedSander:
O I'm not saying that we don't need to find someone to land it. Just I want to find that someone.

matanlurey:
You could put this in @srawlins planned "3.0.0" change?

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: mockito
-version: 3.0.0-alpha
+version: 3.0.0
```
srawlins: Why change this?

grouma: alpha has already been released.

srawlins: True. Let's call it `3.0.0-dev` for now, and not release it. 3.0.0 will be released not super soon, like March probably. I don't want anyone with an itchy trigger finger to publish 3.0.0 till it's ready.

test/mockito_test.dart
```diff
@@ -260,8 +260,8 @@ void main() {
     });
 
     test("should return mock to make simple oneline mocks", () {
-      RealClass mockWithSetup =
-          when(new MockedClass().methodWithoutArgs()).thenReturn("oneline");
+      RealClass mockWithSetup = new MockedClass();
```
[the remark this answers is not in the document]
srawlins: Woohoo!

pubspec.yaml
grouma: Done.

srawlins:
Thanks Gary, love it!

PR #103, opened 2018-02-24
Revert "Update travis script to actually run dartanalyzer"

matanlurey:
Reverts dart-lang/mockito#102

Lets finish landing https://github.com/dart-lang/mockito/pull/100 instead.


kevmoo:
Is it even worth reverting?

matanlurey:
I guess not. I will end up reverting it in #100 anyway.

PR #108, opened 2018-03-23
Bump to 2.2.4 with generic thenReturn/Answer feature

srawlins:
This feature should prevent Open Source projects from landing code that is not allowed internally, until we are able to release 3.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 2.2.2
+
+* Remove last violation of `USES_DYNAMIC_AS_BOTTOM` error. This should make
+  Mockito compatible with Dart 2 semantics.
+
+## 2.2.1
+
+* Internal fixes only (stop using comment-based generic method syntax).
+
 ## 2.2.0
 
 * Add new feature to wait for an interaction: `untilCalled`. See the README for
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.2.4
```
matanlurey: Honestly, this should probably be 3.0.0-alpha, even though it only breaks with Dart2.

We should make the CHANGELOG a bit scarier, i.e. "for early users of Dart2, this is a breaking change". It would be even better to show an example of how to fix a common breakage due to this change.

srawlins: I think this gets into the fuzzy areas of "breaking". It only breaks code that was maybe not doing what authors thought it was doing anyways. It was not documented that you could use Mockito to stub methods with return types other than what the overridden method returns.

The exception to this is probably the case of returning X when the overridden method returns Future<X>, and the value is always `await`ed, so it doesn't matter.

3.0.0-alpha has been released, but we could cut a 3.0.0-beta. or a 3.0.0-alpha-2. WDYT?

matanlurey: I don't know. It effected hundreds of users internally, and Flutter, which means even if we didn't want it to be breaking it is. 3.0.0-alpha+2 SGTM. WDUT?

PR #131, opened 2018-05-14
Improve verify failure message; more tests

### LINKED ISSUE #127: Format `No matching calls` error message better
When no positional args, there is a weird leading comma: `No matching calls. All calls: MockFoo.startDownload(, {request: ..., ...})`

srawlins:
Fixes #127

### DIFF
--- a/lib/src/mock.dart
+++ b/lib/src/mock.dart
@@ -464,7 +464,7 @@ class RealCall {
           .map((key) =>
               "${_symbolToString(key)}: ${invocation.namedArguments[key]}")
           .join(", ");
-      args += ", {$namedArgs}";
+      args += (args.isEmpty) ? "{$namedArgs}" : ", {$namedArgs}";
     }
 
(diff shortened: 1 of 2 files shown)


lib/src/mock.dart
```diff
@@ -464,7 +464,7 @@ class RealCall {
           .map((key) =>
               "${_symbolToString(key)}: ${invocation.namedArguments[key]}")
           .join(", ");
-      args += ", {$namedArgs}";
+      args += (args.isEmpty) ? "{$namedArgs}" : ", {$namedArgs}";
```
TedSander: Nit can you separate this? I almost thought this was the same logic since the comma blends in. I think below would be easier to read, but understand if you want to leave the possibly slightly more performant version.

if (args.isNotEmpty) args += ", ";
args += "{$namedArgs}";

srawlins: Done!

PR #133, opened 2018-05-16
Support nSM Forwarding

### LINKED ISSUE #122: Handle new noSuchMethod mechanics
In Dart 2, it seems likely that NSM will change:

* **Currently**: missing optional arguments are not present in the Invocation object (in `positionalArguments` and `namedArguments`).
* **New**: missing optional arguments _are_ present, and have the default value specified in the parameter.

See https://github.com/dart-lang/sdk/issues/33031.

Mockito then sees all these null values in the Invocation, and assumes that the user forgot to use <code>foo: typed(any<strong>, named: "foo"</strong>)</code>. It requires each named argument to be given a name, so that Mockito can store them properly. How to handle this proposed change?

CC @matanlurey @TedSander

srawlins:
This is the equivalent of #129, for master (3.0.0-beta ish) and fixes #122.

The tests are updated to support both Dart runtimes without NSM Forwarding, and with NSM Forwarding, so that Travis is happy, before and after NSM Forwarding lands for CFE (VM and dart2js), and when it lands for DDC.

### DIFF
--- a/lib/src/mock.dart
+++ b/lib/src/mock.dart
@@ -215,20 +215,10 @@ class _InvocationForMatchedArguments extends Invocation {
     invocation.namedArguments.forEach((name, arg) {
       if (arg == null) {
         if (!_storedNamedArgSymbols.contains(name)) {
-          // Incorrect usage of an ArgMatcher, something like:
-          // `when(obj.fn(a: any))`.
-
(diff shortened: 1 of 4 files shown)


srawlins:
Done.

PR #146, opened 2018-06-21
Upgrade pkg deps; implement workaround in tests; minor code cleanup

### LINKED ISSUE #145: Upgrade dependency on `test`
`test` 1.0.0 has been released, which, according to the changelog:

> No change from 0.12.42.

Maybe just change the upper bound on the dependency constraints:

https://github.com/dart-lang/mockito/blob/78ecd603ce2e52c16db80b786dfb0addad2c7a00/pubspec.yaml#L15

cc @kwalrath @kevmoo

chalin:
Fixes #145 
Works around https://github.com/dart-lang/sdk/issues/33565

- Upgrade dependencies; in particular, support test 1.0.0
- Fix code to eliminate deprecation notices.
- Generalize test helpers to support RegExp's; this is as a workaround to https://github.com/dart-lang/sdk/issues/33565
- Fix .md link

cc @kevmoo @kwalrath

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.0.0-beta+1
+
+- Upgrade package dependencies.
+
 ## 3.0.0-beta
 
 * This release is the first 3.0.0 release featuring the new Mockito 3 API. The
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,7 @@
 Want to contribute? Great! First, read this page (including the small print at the end).
 
 ### Before you contribute
+
 Before we can use your code, you must sign the
 [Google Individual Contributor License Agreement]
 (https://cla.developers.google.com/about/google-individual)
@@ -17,13 +18,14 @@ possibly guide you. Coordinating up front makes it much easier to avoid
 frustration later on.
 
 ### Code reviews
+
 All submissions, including submissions by project members, require review. We
 use GitHub pull requests for this purpose.
 
 ### The small print
+
 Contributions made by corporations are covered by a different agreement than
 the one above, the
-[Software Grant and Corporate Contributor License Agreement]
-(https://cla.developers.google.com/about/google-corporate).
+[Software Grant and Corporate Contributor License Agreement](https://cla.developers.google.com/about/google-corporate).
 
 **NOTE:** This is not an official Google product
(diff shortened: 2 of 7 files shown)


srawlins:
@chalin do you know what the Travis failures are about?

chalin:
Not yet, I'm still investigating.

chalin:
It seems that dart2js now reports errors concerning setters w/o an "=" in the error message. I'm fixing the corresponding tests.

chalin:
Hmm, that isn't it. Investigating further.

chalin:
The cause of the test failures: https://github.com/dart-lang/sdk/issues/33565

srawlins:
Ugh, I dunno if #33565 is going to be prioritized. It might even be intentional. Can you instead:

* adjust the expected failure message on 196 be a RegExp and like, maybe skip the `==A` part, and
* adjust `expectFail` to accept a Pattern instead of a String for `expectedMessage`, and use `expectedMessage.matches(e.message)` on line 65.

chalin:
Sounds like a good plan (the thought had crossed my mind)! Thanks for the pointers, I'll do that.

chalin:
@srawlins - done, and Travis is green. PTAL

chalin:
\o/

Will this be published soon?

srawlins:
Oops. This does _not_ work with Dart ~ 2.0.0-dev.63.0 and earlier:

```
srawlins-macbookair:dart-mockito srawlins$ pub run test
00:01 +37 -1: test/verify_test.dart: verify should mock setter [E]
  NoSuchMethodError: Attempted to use type 'RegExp' as a function. Since types do not define a method 'call', this is not possible. Did you intend to call the
RegExp constructor and forget the 'new' operator?
  Receiver: RegExp
  Tried calling: RegExp("No matching calls\\. All calls: _MockedClass\\.setter=?=A\n\\(If you called `verify\\(\\.\\.\\.\\)\\.called\\(0\\);`, please instead use `verifyNever\\(\\.\\.\\.\\);`\\.\\)")
  dart:core                     Object.noSuchMethod
  test/verify_test.dart 194:35  main.<fn>.<fn>
```
(Optional new)

chalin:
> Oops. This does not work with Dart ~ 2.0.0-dev.63.0 and earlier

Unless `--preview-dart-2` is used.

(It certainly wouldn't hurt at this point to add the optional new. Would you like for me to do that, or simply suggest that folks use `--preview-dart-2` for <2-dev.64?)

chalin:
(Or `export DART_VM_OPTIONS=--preview-dart-2`)

chalin:
Another option is to bump up the min SDK constraint to 2-dev.64.1?

srawlins:
Published!

chalin:
🎉

PR #157, opened 2018-07-14
chore: set max SDK version to <3.0.0

chalin:
cc @kwalrath @kevmoo

### DIFF
--- a/lib/src/mock.dart
+++ b/lib/src/mock.dart
@@ -766,8 +766,8 @@ _InOrderVerification get verifyInOrder {
         if (allInvocations.isNotEmpty) {
           otherCalls = " All calls: ${allInvocations.join(", ")}";
         }
-        fail("Matching call #${tmpVerifyCalls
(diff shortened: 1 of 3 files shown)


lib/src/mock.dart
```diff
@@ -766,8 +766,8 @@ _InOrderVerification get verifyInOrder {
         if (allInvocations.isNotEmpty) {
           otherCalls = " All calls: ${allInvocations.join(", ")}";
         }
-        fail("Matching call #${tmpVerifyCalls
-            .indexOf(verifyCall)} not found.$otherCalls");
+        fail(
+            "Matching call #${tmpVerifyCalls.indexOf(verifyCall)} not found.$otherCalls");
```
srawlins: It looks like this is now longer than 80 chars, can you split the string at...  `} ""not found.`

chalin: Done

srawlins:
Thanks!

PR #159, opened 2018-07-18
Throw an error when stubbing a non-mock method

### LINKED ISSUE #158: Mockito can guard against misleading stubbing of real methods.
Take this mock:

```dart
class MockA extends Mock implements A {
  @override
  int foo() => 7
}
```

If someone tries to stub `foo()`, mockito will pretend to do it, but it won't work:

```dart
var a = new MockA();
when(a.foo()).thenReturn(6);
a.foo(); // returns 7.
```

I think it would be easy for Mockito to crash on the stubbing line above. `noSuchMethod` could return `const WHEN_STUB = const Object()` during a `when` call. Then, in the `when` call, if the method returns something other than`WHEN_STUB`, like `7`, it can throw, saying "Uh... this doesn't look like it was properly stubbed."

WDYT @matanlurey

srawlins:
Fixes #158 

* [x] sanity check internally

### DIFF
--- a/lib/src/mock.dart
+++ b/lib/src/mock.dart
@@ -348,6 +348,10 @@ class PostExpectation<T> {
   }
 
   void _completeWhen(Answering<T> answer) {
+    if (_whenCall == null) {
+      throw new StateError(
+          'Mock method was not called within `when()`. Was a real method called?');
+    }
     _whenCall._setExpected<T>(answer);
     _whenCall = null;
     _whenInProgress = false;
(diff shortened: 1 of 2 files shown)


matanlurey:
Is this (somehow) a breaking change? Do we need to sanity check internally?

srawlins:
I can do that; added a checkbox above.

srawlins:
Sanity check LG.

PR #161, opened 2018-07-23
Bump to 3.0.0 final

srawlins:


kevmoo:
Can we get this out? Need to upgrade angular tests to this to angular tests working with Dart 2 stable

srawlins:
Out.

PR #162, opened 2018-08-06
Add some more capture tests

srawlins:


srawlins:
I don't think this warrants a changelog or version bump

PR #170, opened 2018-10-31
Change dependency from test to test_api

jonahwilliams:
This removes ~20 transitive dependencies of mockito. For more background see: https://github.com/dart-lang/test/pull/945.  Happy to discuss in person - also did not bump version

### DIFF
--- a/lib/src/mock.dart
+++ b/lib/src/mock.dart
@@ -17,7 +17,7 @@ import 'dart:async';
 import 'package:meta/meta.dart';
 import 'package:mockito/src/call_pair.dart';
 import 'package:mockito/src/invocation_matcher.dart';
-import 'package:test/test.dart';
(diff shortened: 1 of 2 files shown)


TedSander:
Like this change. We would want to bump the version right? or I guess that depends if @srawlins  has published recently.

Also just to double check we are bringing test_api internally right?

jonahwilliams:
We should bump the version, just not sure what too - I don't believe the change has landed internally yet.

TedSander:
I'll let @srawlins make the call on the version. No problem about not having the changes internally yet, just want to make sure it will be. Of course it looks like we could still just depend on test for the time being if we needed to sync early for some reason.

natebosch:
Yes `test_api` will be synced internally

srawlins:
I'll bump the version separately.

PR #173, opened 2018-11-05
Roll test_api version to be compatible with test_core

jonahwilliams:
Apparently `^` doesn't grab new minor versions if the package itself is < 1.0


srawlins:
Yeah that's documented here: https://www.dartlang.org/tools/pub/dependencies#caret-syntax. I assume we took this from NPM? Not sure.

PR #176, opened 2018-11-05
Use the new test_api package, checking for version incompatibility

srawlins:
Using 4.0.0 means that Flutter users depending on "^3.0.0" will not upgrade to this release while flutter doesn't support test 1.4.0.

This includes a check for incompatibilities between test and test_api.

A similar 3.0.3 release will be made that does _not_ bump to test_api, but still includes this compatibility check, for a better user experience.

CC @jonahwilliams

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -15,7 +15,7 @@ dependencies:
   collection: ^1.1.0
   matcher: ^0.12.3
   meta: ^1.0.4
-  test_api: ^0.2.0
+  test_api: ^0.2.1
 
 dev_dependencies:
   build_runner: ^1.0.0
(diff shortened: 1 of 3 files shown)


pubspec.yaml
```diff
@@ -15,10 +15,11 @@ dependencies:
   collection: ^1.1.0
   matcher: ^0.12.3
   meta: ^1.0.4
-  test: ^1.2.0
+  test_api: ^0.2.0
```
jonahwilliams: 0.2.1, if it won't get pulled in by the caret

srawlins: Done.

srawlins:
Oops. I guess I need test ^1.5.0. https://travis-ci.org/dart-lang/mockito/jobs/451098672

PR #191, opened 2019-04-30
Add a Fake class with a throwing noSuchMethod

### LINKED ISSUE #169: Add a `Fake` class - noSuchMethod throws
I'm noticing a common pattern is using `extends Mock` but then still manually implementing a bunch of overridden methods and treating the class as a Fake - never using `when()` or `verify()` with it. The `Mock` superclass is only applied because it has a `noSuchMethod` which means that they don't need to add _all_ the overrides, just the ones that are called in the test.

A better practice would be to implement a `Fake` once in a testing library next to the class, but that doesn't seem to be something authors are willing to do. If they are going to use and `extends` to pick up a `noSuchMethod` this way maybe we should give them a supported way to do it that still nudges towards better naming.

Thoughts @srawlins @matanlurey ?

natebosch:
Closes #169

Adds a slightly better pattern for current abuses of the `Mock` class
when it is used only to avoid implementing an entire interface rather
than for it's actual mocking capabilities.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.1.0
+
+* Add a `Fake` class for implementing a subset of a class API as overrides
+  without misusing the `Mock` class.
+
 ## 4.0.0
 
 * Replace the dependency on the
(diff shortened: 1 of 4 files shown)


srawlins:
Great feature! Thanks @natebosch !

natebosch:
@srawlins - any chance we can get this published?

srawlins:
Done!

PR #197, opened 2019-05-31
Fix some typos, it's -> its

natebosch:
It's is a contraction, not possessive.


natebosch:
This was surfaced with an internal lint.

PR #204, opened 2019-07-19
Document anyNamed and captureAnyNamed.

srawlins:


lib/src/mock.dart
```diff
@@ -656,6 +656,12 @@ class _VerifyCall {
       'VerifyCall<mock: $mock, memberName: ${verifyInvocation.memberName}>';
 }
 
+// An argument matcher that acts like an argument during stubbing or
```
natebosch: Triple slash for doc comments

srawlins: This is not a doc comment. This class is package-private, and not exported in the public interface.

natebosch: Private and package private members can be documented too. The doc comment is still useful within the package in your IDE.

srawlins: Is there a way to prevent this class (or its doc comment) from being documented on Pub?

natebosch: The `lib/src` mechanism which makes it package private should prevent it from showing up in the API docs on pub.

srawlins: Ah, good. Thanks!

PR #205, opened 2019-07-25
Improve docs for VerificationResult

srawlins:


lib/src/mock.dart
```diff
@@ -725,15 +726,61 @@ Null _registerMatcher(Matcher matcher, bool capture, {String named}) {
...
+  ///
+  /// Named arguments are listed in the order they are captured in, not the
+  /// order in which they were passed.
+  // TODO(srawlins): make this final, in a major release, as a breaking change.
   List<dynamic> captured = [];
+
+  /// The number of calls matched in this verification.
   int callCount;
 
   // Whether the test API mismatch has been checked.
```
natebosch: [nit] consider removing since it repeats the code?

lib/src/mock.dart
```diff
@@ -725,15 +726,61 @@ Null _registerMatcher(Matcher matcher, bool capture, {String named}) {
...
 
-  VerificationResult(this.callCount) {
-    captured = List<dynamic>.from(_capturedArgs, growable: false);
+  @Deprecated(
+      'User-constructed VerificationResult is deprecated; this constructor may '
+      'be deleted as early as Mockito 5.0.0')
+  VerificationResult(int callCount) : this._(callCount);
+
+  VerificationResult._(this.callCount) {
+    captured = [..._capturedArgs];
```
natebosch: [nit] Consider assigning in the initializer list (which you'd need to do if it was final)

PR #209, opened 2019-08-08
Add docs in README and example/ for Fake.

srawlins:


README.md
```diff
@@ -278,6 +278,36 @@ cat.eatFood("Fish");
 await untilCalled(cat.eatFood(any)); // Completes immediately.
 ```
 
+## Writing a fake
+
+You can also write a simple fake class that implements a real class, by
+extending [Fake]. Fake allows your subclass to satisfy the implementation of
+your real class, without overriding each method with a fake; the Fake class
```
natebosch: how about "without overriding the methods that aren't used in your test"

srawlins: Done. Thanks.

srawlins:
I think I like putting a Best Practices being in README. Its a great one-stop doc now, and I think that would be useful into, and would not bloat the doc.

natebosch:
Sounds good. I'll take a first pass at writing this out and send something your way this afternoon.

PR #216, opened 2019-08-26
Throw an error if an arg matcher is used improperly; 4.1.2

srawlins:
Fixes the confusion that lead to dart-lang/build#4735


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 4.1.2
```
natebosch: I just noticed that `4.1.1` isn't published to pub yet - should we roll these changes in with that an publish?

PR #230, opened 2019-12-06
Fix newly enforced package:pedantic lints

natebosch:
- omit_local_variable_types
- prefer_conditional_assignment
- prefer_if_null_operators
- prefer_single_quotes
- use_function_type_syntax_for_parameters

Also ignore a deprecated import.

Change a check against null to the string "null" to use string
interpolation which already handles `null`.

Remove some inferrable argument types on function literals.

### DIFF
--- a/lib/src/mock.dart
+++ b/lib/src/mock.dart
@@ -17,6 +17,7 @@ import 'dart:async';
 import 'package:meta/meta.dart';
 import 'package:mockito/src/call_pair.dart';
 import 'package:mockito/src/invocation_matcher.dart';
+// ignore: deprecated_member_use
 import 'package:test_api/test_api.dart';
 // TODO(srawlins): Remove this when we no longer need to check for an
 // incompatiblity between test_api and test.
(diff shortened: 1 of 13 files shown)


srawlins:
Oof, so I have not flipped this package to be google3-first yet, while I've made a half-dozen new commits in google3. How should we land this? Should I just flip the copybara config so that we're google3-first, then force you to rebase? I don't think I've changed much existing code; its mostly the new builder file and builder test.

natebosch:
> Should I just flip the copybara config so that we're google3-first, then force you to rebase?

That should be fine. I can also send an internal CL for this and we can wait for a sync to get it here.

lib/src/mock.dart
```diff
@@ -541,8 +543,7 @@ class RealCall {
   @override
   String toString() {
     var argString = '';
-    var args = invocation.positionalArguments
-        .map((v) => v == null ? "null" : v.toString());
+    var args = invocation.positionalArguments.map((v) => '$v');
```
srawlins: Haha, good catch here.

srawlins:
I'd like to just land this now and deal with the divergence internally.

PR #231, opened 2019-12-11
Update README.md example to use Invocation

CaseyHillers:
The following code snippet (as proposed currently in the README) fails:
```
      when(mock.field).thenAnswer(() => boolValue
          ? entryA
          : entryB);
```
`The argument type 'Entry Function()' can't be assigned to the parameter type 'Entry Function(Invocation)`

Adding an `Invocation` parameter makes the code snippet work.
```
      when(mock.field).thenAnswer((Invocation invocation) => boolValue
          ? entryA
          : entryB);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ expect(() => cat.lives, throwsRangeError);
 
 // We can calculate a response at call time.
 var responses = ["Purr", "Meow"];
-when(cat.sound()).thenAnswer((Invocation invocation) => responses.removeAt(0));
+when(cat.sound()).thenAnswer((_) => responses.removeAt(0));
 expect(cat.sound(), "Purr");
 expect(cat.sound(), "Meow");
 ```
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ expect(() => cat.lives, throwsRangeError);
 
 // We can calculate a response at call time.
 var responses = ["Purr", "Meow"];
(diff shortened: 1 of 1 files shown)


jonahwilliams:
cc @srawlins

README.md
```diff
@@ -65,7 +65,7 @@ expect(() => cat.lives, throwsRangeError);
 
 // We can calculate a response at call time.
 var responses = ["Purr", "Meow"];
-when(cat.sound()).thenAnswer(() => responses.removeAt(0));
+when(cat.sound()).thenAnswer((Invocation invocation) => responses.removeAt(0));
```
natebosch:
```suggestion
when(cat.sound()).thenAnswer((_) => responses.removeAt(0));
```

The dart idiom for an argument necessary to fit a signature, but unused, is to name it `_`.

We also don't need the type annotation on `Invocation` because it can be inferred by the signature of the argument to `.thenAnswer`.

CaseyHillers:
@googlebot I fixed it (hopefully)

CaseyHillers:
@natebosch CLA failing is because the suggested commit submitted with your personal gmail.

natebosch:
> CLA failing is because the suggested commit submitted with your personal gmail.

That is strange. It should be fine since it found the CLA for the original one. I think I can merge anyway.

natebosch:
Thanks!

PR #242, opened 2020-03-11
checking on travis runs on different SDKs

kevmoo:


kevmoo:
Sadly, this is showing that we have failures with latest dev build. This is
me testing to see if it really was because of the SDK. I don't think we
should land this. We should figure out why things are feeling with 2.8-dev

kevmoo:
@srawlins – so everything started working. See my SDK bump. Needed to enable testing on lowest supported SDK

PR #254, opened 2020-05-11
Minor tweaks

allentv:
Ran few file formatters and fixed the issues that came up

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
-Mock library for Dart inspired by [Mockito](https://github.com/mockito/mockito).
+# Mock library for Dart inspired by [Mockito](https://github.com/mockito/mockito).
+
+**NOTE:** This is not an official Google product
 
 [![Pub](https://img.shields.io/pub/v/mockito.svg)](https://pub.dartlang.org/packages/mockito)
 [![Build Status](https://travis-ci.org/dart-lang/mockito.svg?branch=master)](https://travis-ci.org/dart-lang/mockito)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -86,9 +88,9 @@ or sometimes when argument matchers are used, etc.
 `ArgumentError`.** This is because it can lead to unexpected behaviors. For
 example:
 
-* If the method is stubbed in a different zone than the zone that consumes the
+- If the method is stubbed in a different zone than the zone that consumes the
```
srawlins, line 91: I prefer to keep these as asterisks.

README.md
```diff
@@ -295,8 +296,8 @@ void main() {
 }
 ```
 
-[Fake]: https://pub.dev/documentation/mockito/latest/mockito/Fake-class.html
-[UnimplementedError]: https://api.dartlang.org/stable/dart-core/UnimplementedError-class.html
+[fake]: https://pub.dev/documentation/mockito/latest/mockito/Fake-class.html
```
srawlins, line 299: I prefer to keep these in the same case as the reference link (lines 275, 277).

lib/src/mock.dart
```diff
@@ -846,7 +846,7 @@ class VerificationResult {
   void _checkTestApiMismatch() {
     try {
       Invoker.current;
-    } on CastError catch (e) {
+    } on TypeError catch (e) {
```
srawlins, line 849: Interesting that this was broken. Can you add a test that fails without this fix?

allentv: This was not broken. It was mentioned as deprecated and the replacement was suggested. I should probably rollback this since it is not a minor tweak and should go with another PR that replaces deprecated features.

allentv:
@srawlins Thanks for the review. I have updated the PR with your comments .

allentv:
@srawlins Thanks for the review. I have updated the PR with your comments .

lib/src/mock.dart
natebosch: We'll also need to ensure that our minimum SDK is at least the one that switched which error type is thrown.

allentv: Shouldn't that be another PR where we fix the SDK versions and the deprecated types together?

natebosch: Yeah another PR would be best, I was calling it out that any PR which does make this change would also need to bump the min SDK.

allentv: I agree. Is someone working on upgrades at the moment? If so, I would like to make a contribution

README.md
```diff
@@ -1,4 +1,6 @@
-Mock library for Dart inspired by [Mockito](https://github.com/mockito/mockito).
+# Mock library for Dart inspired by [Mockito](https://github.com/mockito/mockito)
+
+**NOTE:** This is not an official Google product
```
srawlins, line 3: I would prefer that this text remain at the bottom of the README. This is pretty standard across our open source projects.

lib/src/mock.dart
srawlins: No I don't believe anyone is working on that.

srawlins:
Thanks!

PR #259, opened 2020-06-02
Add any matcher example in README.md

edwinnyawoli:
The other matchers shown here are functions but `any` is not. This may cause users to assume all matchers are functions and try calling `any()`.


edwinnyawoli:
@googlebot I signed it!

srawlins:
Thanks!

PR #270, opened 2020-08-24
Test on stable on Travis

srawlins:


srawlins:
This does not appear to work. :( https://travis-ci.org/github/dart-lang/mockito/jobs/720798157

PR #272, opened 2020-08-27
Remove dep override on build_resolvers

srawlins:
This was to overcome a bug with old-ish http_multi_server and Dart 2.8. It's so far-flung now, it's easier just to remove the dep.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -10,6 +10,9 @@ environment:
 dependencies:
   analyzer: ^0.39.15
   build: ^1.3.0
+  # We don't actually depend directly on build_resolvers, but this version
+  # constraint is important to avoid certain bugs.
+  build_resolvers: ^1.3.10
   code_builder: ^3.4.0
   collection: ^1.1.0
   dart_style: ^1.3.6
@@ -25,6 +28,3 @@ dev_dependencies:
   package_config: ^1.9.3
   pedantic: 1.10.0-nullsafety
   test: ^1.5.1
-
-dependency_overrides:
-  build_resolvers: ^1.3.10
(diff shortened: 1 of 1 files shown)


pubspec.yaml
```diff
@@ -10,6 +10,9 @@ environment:
 dependencies:
   analyzer: ^0.39.15
   build: ^1.3.0
+  # We don't actually depend directly on build_resolvers, but this version
+  # constraint is important to avoid certain bugs.
```
natebosch, line 14: What are the bugs? I would expect that for the most part we'd tell users to upgrade if they hit issues rather than add constraints like this. cc @jakemac53  for thoughts.

Separately when that override was added we stopped testing on the stable SDK. Can we add back stable SDK testing? Can we add testing a `2.7.0`?

srawlins: > What are the bugs?

This has to do with http_multi_server, I think when I was testing whether `pub downgrade` works with the current constraints. build_resolvers 1.3.8 results in http_multi_server 2.1.0 (instead of 2.2.0), which is broken for Dart 2.8.1 ([CHANGELOG entry](https://github.com/dart-lang/sdk/blob/master/CHANGELOG.md#dartio-2)).

> I would expect that for the most part we'd tell users to upgrade if they hit issues rather than add constraints like this. cc @jakemac53 for thoughts.

I guess I could.

> Separately when that override was added we stopped testing on the stable SDK. Can we add back stable SDK testing? Can we add testing a 2.7.0?

Because this depends on test_api 0.2.19-nullsafety, and all of the transitive nullsafety deps, we cannot test on 2.9.0; I think there are version constraints in those deps requiring 2.10.0, or library-level language version comments requiring 2.10.0. So I don't think I can test on 2.7.0 here...

PR #281, opened 2020-09-25
Add an exception builder for throwOnMissingStub.

zhenqiu1101:


natebosch:
I'm going to close this for now since there hasn't been any movement and it doesn't look like a high impact feature. We can revisit with a discussion of the use case if this is still worth considering.

PR #299, opened 2020-11-23
Update collection, matcher, and test deps to nullsafety

felangel:


srawlins:
Mockito's source of truth is internal to Google right now. I'll bump these deps and export to GitHub ASAP.

srawlins:
Ok these have been bumped.

PR #300, opened 2020-11-23
Fix analyzer failures in example/iss

felangel:
- add `pubspec.yaml` to `example` which defines `http` dependency


srawlins:
We've got http as a dev_dependency for this example, so I think we're ok now without this PR. Thanks for opening though!

PR #328, opened 2021-02-02
Migrate to GitHub Actions

athomas:


athomas:
I didn't re-enable coverage, yet. I believe something like this may be needed, but I couldn't get `dart test --coverage` to work.

```
      - name: Collect coverage
        run: dart test --coverage var
      - name: Upload coverage
        uses: coverallsapp/github-action@v1.1.2
        with:
          github-token: ${{ secrets.GITHUB_TOKEN }}
          path-to-lcov: var/lcov.info
```

athomas:
I didn't remove travis.sh in this CL, in case it's useful for local testing.

srawlins:
Thanks @athomas ! I think @jcollins-g figured out coverage for dartdoc, so I'll poke around over there after this lands.

srawlins (approved):

PR #329, opened 2021-02-03
Add a dartdoc verify action in GitHub Actions

srawlins:


srawlins:
CC @athomas @devoncarew This is the result of my experiment to validate docs.

jcollins-g (approved):
Not an expert yet on github actions, but the dartdoc piece definitely looks good to me.

Filed dart-lang/dartdoc#2504 as the flag to set broken-link as an error doesn't seem to take effect here for some reason.

srawlins:
CC @kevmoo I thought you might be interested in this idea too

kevmoo:
We need to put this in a bottle!

Maybe something built off https://github.com/dart-lang/setup-dart – 

CC @mit-mit @athomas

mit-mit:
Rather than adding a complex command like that in setup-dart, I'd much rather get it into `dart` as something like `dart doc --dry-run`.

PR #351, opened 2021-03-09
update to the upcoming test_api

jakemac53:
This needs to be updated prior to publishing test_api so we can roll internally and test that version, then we can publish and remove the override here.


jakemac53:
Closing in favor of landing this internal first and then pushing out externally once test_api is already released

PR #353, opened 2021-03-10
Expand test_api dep to allow test_api 0.3.x

### LINKED ISSUE #352: Provide a non-codegen mockito release with updated test_api dependency.
Currently the flutter_tools project and several dev/ bits of the flutter SDK repo depend on mockito. Because we do not want to introduce a codegen requirement for developing with the SDK (and because we'd like to reduce mocks anyway) we have not been able to upgrade past 4.1.1. We anticipate this migrating will take at least a few more months.

Unfortunately mockito 4.1.1 depends on test_api, which means it constrains our test package version and thus everything that depends on test (shelf, shelf_packages, vm_service, dwds, analyzer).

To avoid being stuck at our current version for the next few months, it would be great to have a release of mockito that supported the latest versions but did not require codegen. Its not necessary for this to be null safe.

SDK version <2.12.0 / test_api ^0.3.0

srawlins:
This is a patch release for mockito 4.1.1.

Fixes #352

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.1.1+1
+
+* Allow test\_api 0.3.x.
+
 ## 4.1.1
 
 * Mark the unexported and accidentally public `setDefaultResponse` as
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: mockito
-version: 4.1.1
+version: 4.1.1+1
 
 authors:
   - Dmitriy Fibulwinter <fibulwinter@gmail.com>
@@ -15,7 +15,7 @@ dependencies:
   collection: ^1.1.0
   matcher: ^0.12.3
   meta: ^1.0.4
-  test_api: ^0.2.1
+  test_api: '>=0.2.1 <0.4.0'
 
 dev_dependencies:
   build_runner: ^1.0.0


srawlins:
CC @jonahwilliams

jonahwilliams (approved):
LGTM

Thank you!

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 4.1.1+1
```
natebosch, line 1:
```suggestion
## 4.1.2
```

Even without code changes we bump the patch version for doc or pubspec changes.

https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#bumping-versions

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: mockito
-version: 4.1.1
+version: 4.1.1+1
```
natebosch, line 2:
```suggestion
version: 4.1.2
```

CHANGELOG.md
srawlins: 4.1.2 has already shipped, and in fact does contain pubspec changes, which @jonahwilliams and flutter_tools do not want.

zanderso:
This PR is needed to resolve a P1 for Flutter, but it looks like it might be stuck. What is the path forward here?

srawlins:
@natebosch is the versioning policy for technical reasons or just idiomatic? I don't know another version number that would be good for this release.

CHANGELOG.md
natebosch: Ah I see.

In that case `4.1.1+1` is OK.

natebosch:
> This PR is needed to resolve a P1 for Flutter

What is the P1 and what other paths to resolution were considered?

zanderso:
> > This PR is needed to resolve a P1 for Flutter
> 
> What is the P1 and what other paths to resolution were considered?

Please see the linked issues https://github.com/flutter/flutter/issues/77856 and https://github.com/dart-lang/mockito/issues/352

PR #362, opened 2021-03-12
Fix 404 link

Goddchen:
The old link just leads to a 404 page. I can't really find the real page. Neither on dart.dev nor on medium. So I linked to the Markdown file on GitHub...


srawlins:
I've noticed this time and time again, and also couldn't find the new link, so I've always given up on changing it. 😛 Thanks for the great fix!

srawlins (approved):

PR #366, opened 2021-03-15
Support latest build packages

### LINKED ISSUE #363: Support latest version of build
I want to upgrade my project to go sound null safe.
It uses `web_socket_channel` and `mockito`. `build_runner` v1.12 depends on `web_socket_channel` v2 but also `build` v2 while `mockito` still depends on `build`v1

Thus this issue is a request to bump the `build` package to v2 in order to enable people upgrading to sound null safety

simolus3:
I raised the `build` dependency to support version 2.x. It looks like we're not using `AssetId.resolve` anywhere, so we don't need to use 2.x as a lower bound.

Closes #363

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.0.2-dev
+
+* Support the latest build packages
+
 ## 5.0.1
 
 * Update to the latest test_api.
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: mockito
-version: 5.0.1
+version: 5.0.2-dev
 
 description: A mock framework inspired by Mockito.
 homepage: https://github.com/dart-lang/mockito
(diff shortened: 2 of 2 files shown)


srawlins (approved):
I love it. Thanks! I'll try to release this tomorrow.

simolus3:
@srawlins Before publishing, I think it would be super awesome to have a coordinated effort between `mockito`, `source_gen` and `build_runner` to support the latest `dart_style` version as well.

cc @jakemac53 Do you know if there were any breaking changes in `dart_style` 2.x for legacy packages? Can we safely use `>=1.0.0 <3.0.0` in `build_runner` and `mockito` that both depend on each other and `dart_style`?

simolus3:
Actually it looks like the packages can update independently by just expanding their versions, sorry for the pings.

srawlins:
No worries; good to have on my radar. On a local checkout, overriding the dart_style dep to 2.0.0 passes all tests, so I'll land a change expanding the constraint to `'>=1.3.6 <3.0.0'` internally, and then export to GitHub.

PR #368, opened 2021-03-16
Support version 2 of dart_style

simolus3:
Increase the upper bound of the `dart_style` dependency to support versions 2.x.

To test this, I manually added a 

```yaml
dependency_overrides:
  dart_style: ^2.0.0
```

All tests and analysis still pass with that override, so I'm pretty confident that we're not affected by any breaking change (the changelog doesn't list any).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 5.0.2-dev
 
 * Support the latest build packages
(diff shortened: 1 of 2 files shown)


simolus3:
Looks like this won't be necessary: https://github.com/dart-lang/mockito/pull/366#issuecomment-800611561

PR #376, opened 2021-03-24
Make verifyInOrder fail given non-mock-calls.

ravenblackx:
Also adds a test for verify failing when given a non-mock call, which is currently untested.

Without this change, this mistake is possible and horrific:

```dart
verifyInOrder([
  fakeService.someCall(),
  fakeService.someOtherCall(),
]);
```

Which just always passes (assuming those calls don't throw), as does e.g. `verifyInOrder(['banana']);`


srawlins (approved):
Looks great, thanks!

PR #385, opened 2021-04-16
Fix some issues with the mockito README.md

jamesderlin:
* The example uses code-generation but doesn't use nullable types. (I
  suppose this might be technically okay, but I'm skeptical that
  anyone would use code-generation without null-safety.)

* Use `dart run` intead of `pub run` to perform code generation.

* Remove references to unstubbed methods returning null by default.

* Fix an example that compared a `List` element to a `List`.

* Fix a broken link.

* Consistently use the American English spelling of "behavior".

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ import 'cat.mocks.dart';
 // Real class
 class Cat {
   String sound() => "Meow";
-  bool eatFood(String food, {bool hungry}) => true;
+  bool eatFood(String food, {bool? hungry}) => true;
   Future<void> chew() async => print("Chewing...");
   int walk(List<String> places) => 7;
   void sleep() {}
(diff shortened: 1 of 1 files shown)


jamesderlin:
The examples in the section with "However, note that `null` cannot be used as an argument adjacent to ArgMatcher arguments..." should be fixed too, but I'm not familiar enough with that to feel comfortable rewriting it.

README.md
```diff
@@ -61,7 +61,7 @@ the Cat class, giving us a class which supports stubbing and verifying.
 
 [NULL_SAFETY_README]: https://github.com/dart-lang/mockito/blob/master/NULL_SAFETY_README.md
 
-## Let's verify some behaviour!
```
srawlins, line 64: This is just the non-American spelling of the word. Not sure which our preference should be...

README.md
```diff
@@ -77,9 +77,6 @@ interactions you are interested in.
 ## How about some stubbing?
 
 ```dart
-// Unstubbed methods return null.
-expect(cat.sound(), nullValue);
```
srawlins, line 81: Good catch.

srawlins (approved):
Thanks!

PR #392, opened 2021-04-25
Fix example tests

CORDEA:
I noticed broken examples, so I want to fix examples.


build.yaml
```diff
@@ -4,6 +4,7 @@ targets:
       mockito|mockBuilder:
         generate_for:
           - test/end2end/*.dart
+          - example/**.dart
```
CORDEA, line 7: Should I change here? If not, I will remove it.

example/example.mocks.dart
```diff
@@ -0,0 +1,64 @@
+// Mocks generated by Mockito 5.0.6 from annotations
```
CORDEA, line 1: Should I commit generated files?

example/example.dart
```diff
@@ -35,6 +40,9 @@ void main() {
   });
 
   test("Let's verify some behaviour!", () {
+    // Stub a method before interacting with it.
+    when(cat.sound()).thenReturn('Meow');
```
CORDEA, line 44: Should I use the `MockCatRelaxed` overall to reduce changes?

build.yaml
srawlins: No that's great. But let's go ahead and alphabetize it, by putting it one line up.

example/example.dart
srawlins: No I love it. Not-relaxed is the new way! 😁

example/example.mocks.dart
srawlins: I have not been checking in the generated test file. You can go ahead and remove this file and I might reconsider later.

example/iss/iss_test.mocks.dart
```diff
@@ -0,0 +1,40 @@
+// Mocks generated by Mockito 5.0.6 from annotations
```
srawlins, line 1: Same here; can remove for now.

srawlins (changes requested):
Thanks so much! Just a few small requests.

CORDEA:
Thank you! I tried to fix it, so please check again.

srawlins (approved):
Sorry, I missed the re-review!

PR #419, opened 2021-05-28
Follow Dart file conventions | Add more detail to the description fie…

fscopel:
Follow Dart file conventions 

Add more detail to the description file description of pubspec.yaml. 
Use 60 to 180 characters to describe the package, what it does, and its target use case.


srawlins (approved):
Thanks!

PR #434, opened 2021-06-21
Ignore lints unnecessary_overrides in generated code

xvrh:
With mockito `v5.0.10` and the commit [Override toString implementation on generated Fakes](https://github.com/dart-lang/mockito/commit/fd5f6039518599bc6050fdf5f1126b2c2bcb314e), the generated code produces a `unnecessary_overrides` lint (from lints/core.yaml)


Tokenyet:
Since this PR is not merged, here is the workarounds I found.

Add the following in `analysis_options.yaml`, and enjoy the clean analyzer.
```
analyzer:
  exclude: [lib/**.g.dart, test/**.mocks.dart]
```

srawlins:
I think this PR is not needed any more, as of mockito 5.0.12, where `toString` is only overridden if additional parameters have been specified.

PR #460, opened 2021-08-03
Fix wrong import for method parameters with default value factories

### LINKED ISSUE #459: Mocking a moor database fails
Generating a mock for a database generated with https://github.com/simolus3/moor fails.
I thought this may be fixed by the new `source_gen` from https://github.com/dart-lang/mockito/issues/425 but this didn't help.

This may happen with all factories that return a subclass which is not declared in the same file.

---

There is a `TableUpdateQuery` class here: https://github.com/simolus3/moor/blob/develop/moor/lib/src/runtime/api/stream_updates.dart#L122 with a factory
`const factory TableUpdateQuery.any() = AnyUpdateQuery;`.

The `AnyUpdateQuery` class is declared in another file here: https://github.com/simolus3/moor/blob/develop/moor/lib/src/runtime/executor/stream_queries.dart#L279

When the mock for the database is generated it contains this fragment for a method where the const factory is used as default value for a parameter:

```dart
import 'package:moor/src/runtime/api/runtime_api.dart' as _i29;
import 'package:moor/src/runtime/executor/stream_queries.dart' as _i32;

  @override
  _i35.Stream<Set<_i29.TableUpdate>> tableUpdates(
          [_i29.TableUpdateQuery? query = const _i32.TableUpdateQuery.any()]) =>
      (super.noSuchMethod(Invocation.method(#tableUpdates, [query]),
              returnValue: Stream<Set<_i29.TableUpdate>>.empty())
          as _i35.Stream<Set<_i29.TableUpdate>>);
```

The `_i32` import prefix is obviously wrong and doesn't point to the file containing the factory function but instead to the type that is being returned by the factory function. This should be `_i29` as well.

@simolus3 FYI

kuhnroyal:
* Fixes problem described in #459 around factory redirects into different source files/imports
* add tests
* gitignore build folder and generated mock test files

Fixes #459

### DIFF
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -1249,7 +1249,8 @@ class _MockClassInfo {
   }
 
   /// Creates a code_builder [Expression] from [object], a constant object from
-  /// analyzer.
+  /// analyzer and [parameter], an optional [ParameterElement], when the
+  /// expression is created for a method parameter.
   ///
   /// This is very similar to Angular's revive code, in
   /// angular_compiler/analyzer/di/injector.dart.
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -1233,8 +1233,9 @@ class _MockClassInfo {
       if (parameter.isNamed) pBuilder.named = true;
       if (parameter.defaultValueCode != null) {
         try {
-          pBuilder.defaultTo =
-              _expressionFromDartObject(parameter.computeConstantValue()!).code;
+          pBuilder.defaultTo = _expressionFromDartObject(
+                  parameter.computeConstantValue()!, parameter)
+              .code;
         } on _ReviveException catch (e) {
           final method = parameter.enclosingElement!;
           final clazz = method.enclosingElement!;
@@ -1252,7 +1253,8 @@ class _MockClassInfo {
   ///
   /// This is very similar to Angular's revive code, in
   /// angular_compiler/analyzer/di/injector.dart.
-  Expression _expressionFromDartObject(DartObject object) {
+  Expression _expressionFromDartObject(DartObject object,
+      [ParameterElement? parameter]) {
     final constant = ConstantReader(object);
     if (constant.isNull) {
       return literalNull;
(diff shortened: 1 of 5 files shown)


srawlins:
I like the idea here, but I think that the parameter element is not sufficient to choosing the right element to reference. For example, here is a test (that you can insert into auto_mocks_test.dart) which does not pass with this fix:

```dart
  test(
      'matches parameter default values constructed from a const factory '
      'constructor2', () async {
    var mocksContent = await buildWithNonNullable({
      ...annotationsAsset,
      ...simpleTestAsset,
      'foo|lib/foo.dart': '''
      import 'baz.dart';
      class Foo {
        void m([Bar a = const Baz.named()]) {}
      }
      class Bar {
        const Bar();
      }
      ''',
      'foo|lib/baz.dart': '''
      import 'foo.dart';
      import 'quux.dart';
      class Baz implements Bar {
        const factory Baz.named() = Quux;
      }
      ''',
      'foo|lib/quux.dart': '''
      import 'baz.dart';
      class Quux implements Baz {
        const Quux();
      }
      ''',
    });
    expect(mocksContent,
        contains('void m([_i2.Bar? a = const _i3.Baz.named()]) =>'));
  });
```

The idea is:

* the parameter comes from one library,
* the factory constructor for the default value comes from a second library,
* the constant value that the factory constructor gives comes from a third library.

I think the fix needs to be made in source_gen.

kuhnroyal:
Yea I didn't think of 3 levels/interface. The information is just not available at that point. I am not sure it would be correct for `computeConstantValue()` to return the type of the factory location. So there probably needs to be something new like `Revivable.type` which can be different from the `DartObject` which was created with `computeConstantValue()`.

I still think the workaround via the `ParameterElement` may be helpful for a fair amount of these cases.

lib/src/builder.dart
```diff
@@ -1247,7 +1248,8 @@ class _MockClassInfo {
   ///
   /// This is very similar to Angular's revive code, in
   /// angular_compiler/analyzer/di/injector.dart.
-  Expression _expressionFromDartObject(DartObject object) {
+  Expression _expressionFromDartObject(DartObject object,
+      [ParameterElement? parameter]) {
```
srawlins, line 1252: Would you mind documenting on the line above, what `parameter` represents?

srawlins (commented):
One comment

lib/src/builder.dart
kuhnroyal: Done

srawlins:
I think your code became unformatted with the sync.

srawlins (approved):

srawlins:
Thanks much! I think this is definitely an improvement (especially with the real world example).

PR #471, opened 2021-09-08
Add ignore_for_file: camel_case_types to generated file

### LINKED ISSUE #463: Add ignore_for_file: camel_case_types
The latest version of the generator (5.0.14) appends a suffix to fake classes (ie: `_FakeMeasure_0`).

It triggers a `Name types using UpperCamelCase. camel_case_types` lint.

RTodorov:
Fixes #463

### DIFF
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -93,7 +93,9 @@ class MockBuilder implements Builder {
           '// ignore_for_file: invalid_use_of_visible_for_testing_member\n'));
       b.body.add(Code('// ignore_for_file: prefer_const_constructors\n'));
       // The code_builder `asA` API unconditionally adds defensive parentheses.
-      b.body.add(Code('// ignore_for_file: unnecessary_parenthesis\n\n'));
+      b.body.add(Code('// ignore_for_file: unnecessary_parenthesis\n'));
+      // The generator appends a suffix to fake classes
+      b.body.add(Code('// ignore_for_file: camel_case_types\n\n'));
       b.body.addAll(mockLibraryInfo.fakeClasses);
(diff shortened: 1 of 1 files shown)


RTodorov:
@googlebot I signed it!

lib/src/builder.dart
```diff
@@ -93,7 +93,9 @@ class MockBuilder implements Builder {
           '// ignore_for_file: invalid_use_of_visible_for_testing_member\n'));
       b.body.add(Code('// ignore_for_file: prefer_const_constructors\n'));
       // The code_builder `asA` API unconditionally adds defensive parentheses.
-      b.body.add(Code('// ignore_for_file: unnecessary_parenthesis\n\n'));
+      b.body.add(Code('// ignore_for_file: unnecessary_parenthesis\n'));
```
kevmoo, line 96: An aside for @srawlins – wonder if we should pull these ignores into a constant and DRY up this logic a bit...

srawlins: I do like having a comment for why each one might be needed.

kevmoo: That could go in the constant definition, right?

srawlins: Yeah, it could.

 I do hope to elimitate all of these and replace with https://github.com/dart-lang/sdk/issues/46957.

RTodorov: In the meantime, could we perhaps get this one merged? :)

srawlins: Yeah, the support just landed for "all lints", but it will be introduced in Dart 2.15, which is many moons from now.

srawlins (approved):

Issue #461, opened 2021-08-03
Please add   // ignore_for_file: unnecessary_overrides to generated files or do not generate toString() overrides


idrats:
Because 
```dart
  @override
  String toString() => super.toString();
```
is not allowed by strict analyser options

woelmer:
5.0.12 does not have the lint warnings. 5.0.13 does

brizaldi:
Hi, I'm currently using `mockito: ^5.0.14` this issue still appears.

`my_test.dart`
```dart
import 'package:connectivity_plus/connectivity_plus.dart';
import 'package:mockito/annotations.dart';

import 'my_test.mocks.dart';

@GenerateMocks([Connectivity])
void main() {
  final mockConnectivity = MockConnectivity();
}
```

`my_test_mocks.dart`
```dart
// Mocks generated by Mockito 5.0.14 from annotations
// in wallpaper_app/test/extra/network/my_test.dart.
// Do not manually edit this file.

import 'dart:async' as _i3;

import 'package:connectivity_plus/connectivity_plus.dart' as _i2;
import 'package:connectivity_plus_platform_interface/connectivity_plus_platform_interface.dart'
    as _i4;
import 'package:mockito/mockito.dart' as _i1;

// ignore_for_file: avoid_redundant_argument_values
// ignore_for_file: avoid_setters_without_getters
// ignore_for_file: comment_references
// ignore_for_file: implementation_imports
// ignore_for_file: invalid_use_of_visible_for_testing_member
// ignore_for_file: prefer_const_constructors
// ignore_for_file: unnecessary_parenthesis

/// A class which mocks [Connectivity].
///
/// See the documentation for Mockito's code generation for more information.
class MockConnectivity extends _i1.Mock implements _i2.Connectivity {
  MockConnectivity() {
    _i1.throwOnMissingStub(this);
  }

  @override
  _i3.Stream<_i4.ConnectivityResult> get onConnectivityChanged =>
      (super.noSuchMethod(Invocation.getter(#onConnectivityChanged),
              returnValue: Stream<_i4.ConnectivityResult>.empty())
          as _i3.Stream<_i4.ConnectivityResult>);
  @override
  _i3.Future<_i4.ConnectivityResult> checkConnectivity() => (super.noSuchMethod(
          Invocation.method(#checkConnectivity, []),
          returnValue:
              Future<_i4.ConnectivityResult>.value(_i4.ConnectivityResult.wifi))
[...]

PR #478, opened 2021-09-23
fix(linter): ignore some linters triggered by generated files

jhionan:
Some generated code is triggering some linters making pipelines with this linters to fail.
Generated files should not trigger any linter.

image

### DIFF
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -98,6 +98,7 @@ class MockBuilder implements Builder {
       b.body.add(Code('// ignore_for_file: camel_case_types\n'));
       b.body.add(Code('// ignore_for_file: always_specify_types\n'));
       b.body.add(Code('// ignore_for_file: unnecessary_overrides\n'));
+      b.body.add(Code('// ignore_for_file: implementation_imports\n'));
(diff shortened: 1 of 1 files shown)


srawlins:
Thanks for the PR.

It is unsustainable to add an ignore line for every lint rule. I've added the common ones and the ones rules used by `lints` and `flutter_lints`, but probably will not add any for other rules.

* always_specify_types - Is this rule recommended somewhere?
* unnecessary_overrides - mockito should not be generating unnecessary overrides. I have a change in flight that fixes #461 differently.
* implementation_imports - This is already ignored on line 88.
* avoid_implementing_value_types - Is this rule recommended somewhere?

jhionan:
avoid_implementing_value_types: was recommended on dart, but they removed this rule when moving to null safety

jhionan:
removed the duplicate `implementation_imports` entry

jhionan:
I'll drop this got the conclusion that the best solution is excluding mocks from anyproject analyzer
```
exclude:
    - "**/*.mocks.dart"
```

srawlins:
Haha sorry about that. I think that's a fine solution for now. In the future we will `// ignore_for_file: type=lint`.

PR #497, opened 2021-12-21
Add ignore all lints comment to generated code

Pante:
This PR introduces a trivial enhancement in the form of the new `ignore_for_file: type=lint` comment that was added in Dart 2.15. It ensures that dart analyzer will ignore the generated file without having to explicitly exclude it in a project's analysis_options.yaml. I'm not sure if any addition tests are needed for this change.

It is a follow-up from https://github.com/dart-lang/sdk/issues/47965.


srawlins (approved):
Love it thanks!

PR #517, opened 2022-02-15
fix: version number in readme

AliAkberAakash:
Updated version number from 5.0.0 to 5.1.0 in [README.md](https://github.com/dart-lang/mockito#lets-create-mocks)


srawlins:
Oops, on second thought, this is intentional. The document is noting the _first_ version of mockito which supports null safety.

srawlins (changes requested):
I think this can be closed.

PR #530, opened 2022-04-18
Update pubspec.yaml

devoncarew:
- add a `repository` field to the pubspec


srawlins:
The source-of-truth for mockito is google3 (even for this file), so updating it there would require one fewer import CL.

devoncarew:
Ah! I'll re-create this as a critique CL.

PR #537, opened 2022-06-04
Fix ISS example readme links.

kasipavankumar:
"Files" section in ISS example's README file points to an invalid (404) location. This PR fixes it by pointing to the correct URL of files.


srawlins (approved):
Thanks!

PR #544, opened 2022-07-10
Minor grammar fix in the readme

NamanShergill:
I wasn't sure if I should have just let it be or opened an issue/PR for such a minor thing but just in case, here's a PR


srawlins (approved):
A PR is great, thanks much for catching that!

PR #582, opened 2022-11-03
Add test case to reproduce issue with throwing on missing stub when a function returns a Future

Barabas5532:
Relates to https://github.com/dart-lang/mockito/issues/581

### DIFF
--- a/test/end2end/foo.dart
+++ b/test/end2end/foo.dart
@@ -19,9 +19,6 @@ class Foo<T> {
   set setter(int? value) {}
   void returnsVoid() {}
   Future<void> returnsFutureVoid() => Future.value();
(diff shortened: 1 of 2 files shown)


Barabas5532:
The test failure message is:

```
00:00 +16 -1: for a generated mock, an unstubbed method returning a future throws [E]
  Expected: throws <Instance of 'MissingStubError'> with `toString()`: contains 'returnsFutureVoid()'
    Actual: <Closure: () => Future<void>>
     Which: returned a Future that emitted <null>

  dart:async  _CustomZone.runUnary
```

Barabas5532:
Closing this, as there are conflicting existing tests, and it is clear that this is the desired behaviour.

PR #592, opened 2022-12-11
support mocking non-nullable generic return types now. 

zhanglix:
Hi there, 

I implemented a version to support mocking members/getters non-nullable generic return types. Since I am new to dart, I am not sure if my implementation comply with this project's convention.

There are about 10 tests failed, which I think are expected since all of them are expecting failure or throw errors on mocking non-nullable methods. 

In my opinion, since we can support non-nullable return types now, a lots of related checking could be removed to make codebase simpler. 

#589

### DIFF
--- a/lib/mockito.dart
+++ b/lib/mockito.dart
@@ -53,6 +53,6 @@ export 'src/mock.dart'
         MissingStubError,
         FakeUsedError,
 
-        //to support non-nullable return types
+        // -- dummy instance creation
         defaultDummyValue,
         setDefaultDummyValue;
--- a/lib/mockito.dart
+++ b/lib/mockito.dart
@@ -51,4 +51,8 @@ export 'src/mock.dart'
         logInvocations,
         untilCalled,
         MissingStubError,
-        FakeUsedError;
+        FakeUsedError,
+
+        //to support non-nullable return types
+        defaultDummyValue,
+        setDefaultDummyValue;
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -821,9 +821,6 @@ class _MockTargetGatherer {
   /// A method is not valid for stubbing if:
   /// - It has a private type anywhere in its signature; Mockito cannot override
   ///   such a method.
-  /// - It has a non-nullable type variable return type, for example `T m<T>()`,
-  ///   and no corresponding dummy generator. Mockito cannot generate its own
-  ///   dummy return values for unknown types.
   void _checkMethodsToStubAreValid(_MockTarget mockTarget) {
     final interfaceElement = mockTarget.interfaceElement;
     final className = interfaceElement.name;
@@ -910,16 +907,6 @@ class _MockTargetGatherer {
       errorMessages.addAll(_checkFunction(returnType, enclosingElement,
           allowUnsupportedMember: allowUnsupportedMember,
           hasDummyGenerator: hasDummyGenerator));
-    } else if (returnType is analyzer.TypeParameterType) {
-      if (!isParameter &&
-          !allowUnsupportedMember &&
-          !hasDummyGenerator &&
-          _entryLib.typeSystem.isPotentiallyNonNullable(returnType)) {
-        errorMessages
-            .add('${enclosingElement.fullName} features a non-nullable unknown '
-                'return type, and cannot be stubbed. '
-                '$_tryUnsupportedMembersMessage');
-      }
     }
 
     for (var parameter in function.parameters) {
@@ -1365,16 +1352,11 @@ class _MockClassInfo {
       return;
     }
 
-    final returnTypeIsTypeVariable =
-        typeSystem.isPotentiallyNonNullable(returnType) &&
-            returnType is analyzer.TypeParameterType;
     final fallbackGenerator = fallbackGenerators[method.name];
     final parametersContainPrivateName =
         method.parameters.any((p) => p.type.containsPrivateName);
     final throwsUnsupported = fallbackGenerator == null &&
-        (returnTypeIsTypeVariable ||
-            returnType.containsPrivateName ||
-            parametersContainPrivateName);
+        (returnType.containsPrivateName || parametersContainPrivateName);
 
     if (throwsUnsupported) {
       if (!mockTarget.unsupportedMembers.contains(name)) {
@@ -1382,8 +1364,7 @@ class _MockClassInfo {
         // [_MockTargetGatherer._checkFunction].
         throw InvalidMockitoAnnotationException(
             "Mockito cannot generate a valid override for '$name', as it has a "
-            'non-nullable unknown return type or a private type in its '
-            'signature.');
+            'private type in its signature.');
       }
       builder.body = refer('UnsupportedError')
           .call([
@@ -1467,6 +1448,12 @@ class _MockClassInfo {
     }
 
     if (type is! analyzer.InterfaceType) {
+      if (type is analyzer.TypeParameterType) {
+        return referImported(
+                'defaultDummyValue<$type>', 'package:mockito/mockito.dart')
+            .call([]);
+      }
+
       // TODO(srawlins): This case is not known.
       return literalNull;
     }
@@ -1482,22 +1469,13 @@ class _MockClassInfo {
       final typeArgument = typeArguments.first;
       final typeArgumentIsPotentiallyNonNullable =
           typeSystem.isPotentiallyNonNullable(typeArgument);
-      if (typeArgument is analyzer.TypeParameterType &&
-          typeArgumentIsPotentiallyNonNullable) {
-        // We cannot create a valid Future for this unknown, potentially
-        // non-nullable type, so we'll use a `_FakeFuture`, which will throw
-        // if awaited.
-        var futureType = typeProvider.futureType(typeArguments.first);
-        return _dummyValueImplementing(futureType, invocation);
-      } else {
-        // Create a real Future with a legal value, via [Future.value].
-        final futureValueArguments = typeArgumentIsPotentiallyNonNullable
-            ? [_dummyValue(typeArgument, invocation)]
-            : <Expression>[];
-        return _futureReference(_typeReference(typeArgument))
-            .property('value')
-            .call(futureValueArguments);
-      }
+      // Create a real Future with a legal value, via [Future.value].
+      final futureValueArguments = typeArgumentIsPotentiallyNonNullable
+          ? [_dummyValue(typeArgument, invocation)]
+          : <Expression>[];
+      return _futureReference(_typeReference(typeArgument))
+          .property('value')
+          .call(futureValueArguments);
     } else if (type.isDartCoreInt) {
       return literalNum(0);
     } else if (type.isDartCoreIterable || type.isDartCoreList) {
@@ -1866,18 +1844,15 @@ class _MockClassInfo {
 
     final returnType = getter.returnType;
     final fallbackGenerator = fallbackGenerators[getter.name];
-    final returnTypeIsTypeVariable =
-        typeSystem.isPotentiallyNonNullable(returnType) &&
-            returnType is analyzer.TypeParameterType;
-    final throwsUnsupported = fallbackGenerator == null &&
-        (returnTypeIsTypeVariable || getter.returnType.containsPrivateName);
+    final throwsUnsupported =
+        fallbackGenerator == null && getter.returnType.containsPrivateName;
     if (throwsUnsupported) {
       if (!mockTarget.unsupportedMembers.contains(getter.name)) {
         // We shouldn't get here as this is guarded against in
         // [_MockTargetGatherer._checkFunction].
         throw InvalidMockitoAnnotationException(
             "Mockito cannot generate a valid override for '${getter.name}', as "
-            'it has a non-nullable unknown type or a private type.');
+            'it has a private type.');
       }
       builder.body = refer('UnsupportedError')
           .call([
--- a/lib/src/mock.dart
+++ b/lib/src/mock.dart
@@ -1268,7 +1268,7 @@ extension ListOfVerificationResult on List<VerificationResult> {
   List<List<dynamic>> get captured => [...map((result) => result.captured)];
 }
 
-final _defaultDummyValueFor = <Type, dynamic>{
+final _defaultDummyValueFor = <Type, Object?>{
   num: 0,
   int: 0,
   double: 0.0,
--- a/lib/src/mock.dart
+++ b/lib/src/mock.dart
@@ -1267,3 +1267,27 @@ extension ListOfVerificationResult on List<VerificationResult> {
   /// [verifyInOrder].
   List<List<dynamic>> get captured => [...map((result) => result.captured)];
 }
+
+final _defaultDummyValueFor = <Type, dynamic>{
+  num: 0,
+  int: 0,
+  double: 0.0,
+  bool: false,
+  String: '',
+};
+
+void setDefaultDummyValue<T>(T value) {
+  _defaultDummyValueFor[T] = value;
+}
+
+T defaultDummyValue<T>() {
+  if (null is T) {
+    return null as T;
+  }
+  if (_defaultDummyValueFor.containsKey(T)) {
+    return _defaultDummyValueFor[T] as T;
+  } else {
+    throw UnsupportedError('No default dummy value specified for $T, '
+        'call setDefaultDummyValue<$T>(value) to set one');
+  }
+}
--- a/lib/src/mock.dart
+++ b/lib/src/mock.dart
@@ -1287,7 +1287,7 @@ T defaultDummyValue<T>() {
   if (_defaultDummyValueFor.containsKey(T)) {
     return _defaultDummyValueFor[T] as T;
   } else {
-    throw UnsupportedError('No default dummy value specified for $T, '
-        'call setDefaultDummyValue<$T>(value) to set one');
+    throw UnsupportedError("No default dummy value specified for '$T', "
+        'call `setDefaultDummyValue<$T>(value)` to set one');
   }
 }
--- a/test/builder/auto_mocks_test.dart
+++ b/test/builder/auto_mocks_test.dart
@@ -3050,7 +3050,7 @@ void main() {
   });
 
   test(
-      'uses defaultDummyValue, given a a class with a method with a '
+      'uses defaultDummyValue, given a class with a method with a '
       'non-nullable class-declared type variable return type', () async {
     await expectSingleNonNullableOutput(
       dedent('''
--- a/test/builder/auto_mocks_test.dart
+++ b/test/builder/auto_mocks_test.dart
@@ -1576,7 +1576,7 @@ void main() {
           FutureOr<R> m<R>();
         }
         '''),
-      _containsAllOf('_i2.FutureOr<R> m<R>() => (super.noSuchMethod('),
+      _containsAllOf('_i3.FutureOr<R> m<R>() => (super.noSuchMethod('),
     );
   });
 
@@ -2310,7 +2310,7 @@ void main() {
         Future<T> m<T>() async => false;
       }
       '''),
-      _containsAllOf('returnValue: _FakeFuture_0<T>('),
+      _containsAllOf('returnValue: _i3.Future<T>.value(_i1.defaultDummyValue'),
     );
   });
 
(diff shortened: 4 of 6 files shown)


srawlins:
CC @yanok

yanok:
First of all, thanks for your PR! I will first reply with some high level comments, will comment on the code later.

Second, if you are new to Dart and don't have a big code base (so can choose the mocking library easily), consider also looking at https://github.com/felangel/mocktail, as it does pretty much what you are trying to implement here + it actually does it without codegen (at the cost of slightly different API), and I don't think we realistically can ditch the codegen in Mockito.

Some high-level points, if you still want to stick with Mockito ;) :
1. We should think how this functionality interacts with `fallbackGenerators`: it seems like they shouldn't be necessary anymore, but people already use them, so we need a migration path. There is also `unsupportedMembers` but I think we can just drop them.
2. Since codegen is going to stay, it would be nice to populate the fallback values map with all the generated fake classes.
3. I have a bit of concern about exposing `setDefaultDummyValue` to the users: in my experience it can and will be abused to massively set expectations and I believe it will have a negative effect on the test readability. Can we add a named argument to `@GenerateNiceMocks` instead, that will take a list of classes, generate fakes for them and insert `origClass` -> `fakeObject` mappings into the fallback map?

zhanglix:
> First of all, thanks for your PR! I will first reply with some high level comments, will comment on the code later.
> 
> Second, if you are new to Dart and don't have a big code base (so can choose the mocking library easily), consider also looking at https://github.com/felangel/mocktail, as it does pretty much what you are trying to implement here + it actually does it without codegen (at the cost of slightly different API), and I don't think we realistically can ditch the codegen in Mockito.

I do have considered using mocktail, but decided to stick to mockito since I think this project definitely have more resources to maintain and improve in the future. It's easy to write some open source project once, but it is really costly to maintain it.

> Some high-level points, if you still want to stick with Mockito ;) :
> 
> 1. We should think how this functionality interacts with `fallbackGenerators`: it seems like they shouldn't be necessary anymore, but people already use them, so we need a migration path. There is also `unsupportedMembers` but I think we can just drop them.

The `fallbackGenerators` is intact for this PR, it's still functional.  Personally, I think it is tedious to specify a fallbackGenerator for each method if you are testing a generic class which has lots of method return a value of a generic type, which is my current situation.


> 2. Since codegen is going to stay, it would be nice to populate the fallback values map with all the generated fake classes.

If the dart-team will:

1.  provide return-type parameter for the Object.noSuchMethod(), and 
2.  extend generic class/mixin definition to support extend/implements a generic type, such as 
   ```dart
        class Mock<T> implements T {} //no code gen any more
        class Fake<T> implements T {} //no defaultDummyValue() any more
        class Any<T> implements T{}//all argument matcher can be implemented this way; no xxxNamed(...) matcher any more.
   ```
we can abandon code-gen finally.

In my opinion, `providing return-type` is a highly possible.
And `expand generic class/mixin definition` is harder but not impossible.  First, its a nature extension of generic programing not a hacking; Second, as a side effect it regain all the flexibility before null-safety enforcement without any compromise of the benefits of null-safety; Third, from the point view of a compiler, it has no fundamental different from a definition like this: 
   ```dart
   class MyList<T> implements List<T> {}
   ```

Further more if the proposal of @lrhn's of invoke can be extended to invoke the same method on another object with the same arguments, like
``` dart
class Invocation<R> {
   ...
   R invoke(X Function<X>(Invocation<X>) handler) => handler<R>(this);

   R invokeOn(Object obj);
}
```
then we can implement Spy too, which is a really useful feature of javascript's mockito. (well, this may conflicts with dart's tree shaking optimization)
```dart
class Spy<T> implements T {
    final T obj
    Spy(this.obj);
    dynamic noSuchMethod(...){
        //...do some thing
        var ret = invocation.invokeOn(obj);
        //...do some thin
        return ret;
    }
}
```

So I think we should propose 1 and 2 to dart team, what if they accept it some day?

> 3. I have a bit of concern about exposing `setDefaultDummyValue` to the users: in my experience it can and will be abused to massively set expectations and I believe it will have a negative effect on the test readability. Can we add a named argument to `@GenerateNiceMocks` instead, that will take a list of classes, generate fakes for them and insert `origClass` -> `fakeObject` mappings into the fallback map?

I don't either want user to abuse `setDefaultDummyValue`. Its usage can be reduced by making `defaultDummyValue<T>()` more powerful so that most users do not need to use it. For example, we can handle those common cases such as primitive types, common container(List/Set/Map/Iterator), stream, future, or even function. (I think the code gen has already dealt with Future and Function quite well).

I think the `origClass` -> `fakeObject` mappings is a great idea, especially for those big classes. But I don't think puts them in @GenerateNiceMocks is a good idea. Personally, I think @GenerateNiceMocks and generate fake object for a type should NOT be bounded together, especially in case of mocking generic classes such as ```MockList<T>```. The mock class should be able generated before the specific `T` is determined. Maybe we can provide a annotation such as `@GenerateFakes` or `@generateFakeValues` which wil do the job of populating dummyValueMap,

Anyway I wish mockito can ditch code-gen, but I am fine with it. Although I think it make mockito a little more complex than needed, which may introduce bugs and harder to maintain. for example, with my PR, if you generate mock like this
```dart
@GenerateNiceMocks([
  MockSpec<MyIterable>(),
])
abstract class MyIterable<T> extends Iterable<T> {}
```
the generated `map` method is not a valid override due to the name collision of parameter type between `MyIterator<T>` and `Iterable.map<T>`. 
```dart
 @override
  Iterable<T> map<T>(T Function(T)? toElement) => (super.noSuchMethod(
        Invocation.method(
          #map,
          [toElement],
        ),
        returnValue: <T>[],
        returnValueForMissingStub: <T>[],
      ) as Iterable<T>);
```

although it can be workaround by rename the type parameter of  `MyIterator` from `T` to `E`, which will generate the correct version of map
```dart
  @override
  Iterable<T> map<T>(T Function(E)? toElement) => (super.noSuchMethod(
        Invocation.method(
          #map,
          [toElement],
        ),
        returnValue: <T>[],
        returnValueForMissingStub: <T>[],
      ) as Iterable<T>);
```


For my current situation `fallbackGenerators` is too tedious since there are too many generic class return value of type `T` in my project, most of those `T`s are primitive types, so I hope some mechanism similar to `setDefaultDummyValue` can make the procedure of mocking those classes as simple as possible. 

And I understand you guys need to consider more general situations than mine, and have your own priorities, I can wait until you figure out the best way to resolve the problem and stick to my patch for my personal usage for now.

yanok:
> The fallbackGenerators is intact for this PR, it's still functional. Personally, I think it is tedious to specify a fallbackGenerator for each method if you are testing a generic class which has lots of method return a value of a generic type, which is my current situation.

BTW, unrelated to this PR. If your problem is mocking generic classes, do you know you can generate instantiated mocks? Since you always know what the generic argument will be used in the tests, you can just get a mock of instantiated class instead, and that won't require `fallbackGenerators`. So `@GenerateNiceMocks([MockSpec<Foo<int>>()])` can remove a lot of pain points (I totally agree that it is worse than just mocking `Foo` and passing a type argument to `MockFoo`, but it is still much better than `fallbackGenerators`).
The real problem are methods that are polymorphic themselves (i.e. their return type is not a type variable coming from the class definition, but a type variable coming from the method definition itself), these really require `fallbackGenerators` today, there is no way out of it.

My point was that, even though your PR doesn't affect `fallbackGenerators`, we would like to migrate off fallback generators in future and just tell people to install dummy values instead, so we need to think about the migration path. But ok, I guess we can discuss this later.

> So I think we should propose 1 and 2 to dart team, what if they accept it some day?

I replied on the dart-language bug. We discussed this before (unfortunately the discussion happened internally, so I can't give you a reference) and the outcome was it's not gonna happen.

> I don't either want user to abuse setDefaultDummyValue

My point is: if you expose it, users **will** abuse it ;) So I want to consider our options to hide it.

> I think the origClass -> fakeObject mappings is a great idea, especially for those big classes. But I don't think puts them in @GenerateNiceMocks is a good idea.

What I meant is `@GenerateNiceMocks` already creates a lot of fakes, we should put them into `_defaultValueFor` map, since we will get these types covered for free. Of course that won't be enough for all cases, so we need to give the user a way to add something more to the map. But I'm not convinced `setDefaultDummyValue` is the API we want to expose. Ideally, I'd like to skip the value part, so a user just tells "I want this type to have a dummy value too" but the library decides on the specific value (a fake ideally, except for the types that can't be faked). This can't be done at runtime, so requires a code generation. But we are already doing code generation, so that's not such a big deal. Yep, that can totally be a separate annotation, doesn't have to be in `@GenerateNiceMocks`.

> The fallbackGenerators is intact for this PR, it's still functional. Personally, I think it is tedious to specify a fallbackGenerator for each method if you are testing a generic class which has lots of method return a value of a generic type, which is my current situation.

Uh, that's bad, could you please file a bug for that? Yep, code generation is hard to done right, unfortunately.

lib/mockito.dart
```diff
@@ -51,4 +51,8 @@ export 'src/mock.dart'
         logInvocations,
         untilCalled,
         MissingStubError,
-        FakeUsedError;
+        FakeUsedError,
+
+        //to support non-nullable return types
+        defaultDummyValue,
```
yanok, line 57: does this need to be public?

lib/src/builder.dart
```diff
@@ -1420,6 +1401,13 @@ class _MockClassInfo {
     }
 
     if (type is! analyzer.InterfaceType) {
+      if (type is analyzer.TypeParameterType) {
+        // return CodeExpression(Code('_i1.defaultDummyValue<$type>()'));
```
yanok, line 1405: please remove

lib/mockito.dart
yanok: Oh, I see, you need to call it from generated code...

lib/src/mock.dart
```diff
@@ -1242,3 +1242,27 @@ extension ListOfVerificationResult on List<VerificationResult> {
...
   List<List<dynamic>> get captured => [...map((result) => result.captured)];
 }
+
+final _defaultDummyValueFor = <Type, dynamic>{
+  num: 0,
+  int: 0,
+  double: 0.0,
+  bool: false,
+  String: '',
+};
```
yanok, line 1252: Ideally I'd like these to automatically agree with the dummy values created by the codegen, so maybe codegen should just generate calls to `defaultDummyValue`... but that's not a big deal.

test/builder/auto_mocks_test.dart
```diff
@@ -1562,19 +1562,22 @@ void main() {
...
-        abstract class Foo {
-          FutureOr<R> m<R>();
-        }
-        '''),
-      _containsAllOf('_i2.FutureOr<R> m<R>() => (super.noSuchMethod('),
-    );
-  });
+  // TODO: please review and verify I can delete this test
+  //I think this is not not expected any more since we support non-nullable
+  // return types now.
```
yanok, line 1567: I don't get why you'd like to remove this test, it wasn't checking for failure, so I don't buy the explanation that it used to fail and now it works and that's what makes the test fail. What's the error you get?

test/builder/auto_mocks_test.dart
```diff
@@ -2298,17 +2301,20 @@ void main() {
...
-      class Foo {
-        Future<T> m<T>() async => false;
-      }
-      '''),
-      _containsAllOf('returnValue: _FakeFuture_0<T>('),
-    );
-  });
+  // TODO: please review and verify I can delete this test
+  //I think this is not not expected any more since we support non-nullable
+  // return types now.
```
yanok, line 2306: Same here, this just needs to be fixed, since we are generating `FakeFuture` with your PR.

test/builder/auto_mocks_test.dart
```diff
@@ -3032,82 +3038,86 @@ void main() {
...
-        '''),
-      },
-      message: contains(
-          "The method 'Foo.m' features a non-nullable unknown return type, and "
-          'cannot be stubbed'),
-    );
-  });
+  // TODO: please review and verify I can delete this test
+  //I think this is not not expected any more since we support non-nullable
+  // return types now.
```
yanok, line 3043: Please turn the "check that it fails" test into "check that it works" and verify that a call to `defaultDummyValue` is generated.

yanok: and same for other tests that used to fail, but work now.

zhanglix:
just synced the changes which made several additionaly commits

zhanglix:
Any update on this PR? @yanok

yanok:
Sorry, I was busy with other stuff.  I'm not sure we support merges, @srawlins do we? Could you maybe rebase and squash your changes? Otherwise it's a bit hard to check how the tests are updated.

I'd love to also have a generated code that populates the map with created fake classes. Seems like a free win.

zhanglix:
> Sorry, I was busy with other stuff. I'm not sure we support merges, @srawlins do we? Could you maybe rebase and squash your changes? Otherwise it's a bit hard to check how the tests are updated.

I am not sure how to do this with github, it seems that github fork update make a merge automatically?

> I'd love to also have a generated code that populates the map with created fake classes. Seems like a free win.

could you please just checkout the code and add this? And resolve the rebase issue btw?

zhanglix:
@yanok any update?

lib/src/mock.dart
```diff
@@ -1270,3 +1270,27 @@ extension ListOfVerificationResult on List<VerificationResult> {
   /// [verifyInOrder].
   List<List<dynamic>> get captured => [...map((result) => result.captured)];
 }
+
+final _defaultDummyValueFor = <Type, dynamic>{
```
srawlins, line 1274: I believe `resetMockitoState` should be updated to clear this Map.

lib/src/mock.dart
```diff
@@ -1270,3 +1270,27 @@ extension ListOfVerificationResult on List<VerificationResult> {
...
+  double: 0.0,
+  bool: false,
+  String: '',
+};
+
+void setDefaultDummyValue<T>(T value) {
+  _defaultDummyValueFor[T] = value;
+}
+
+T defaultDummyValue<T>() {
```
srawlins, line 1286: To prevent accidentally calling `defaultDummyValue()` (without an explicit type argument), I think we should throw if `T` is `dynamic`.

lib/src/mock.dart
```diff
@@ -1270,3 +1270,27 @@ extension ListOfVerificationResult on List<VerificationResult> {
...
+
+final _defaultDummyValueFor = <Type, dynamic>{
+  num: 0,
+  int: 0,
+  double: 0.0,
+  bool: false,
+  String: '',
+};
+
+void setDefaultDummyValue<T>(T value) {
```
srawlins, line 1282: To prevent accidentally calling `defaultDummyValue()` (without an explicit type argument), I think we should throw if `T` is `dynamic`. (Or, if a bound is added, maybe throw if `T` is `Object`?)

test/end2end/generated_mocks_test.dart
```diff
@@ -229,6 +260,21 @@ void main() {
...
       baz.returnsTypeVariable();
     });
+
+    test(
+      'access typeVariableField will throw UnsupportedError before '
+      'setting default via setDefaultDummyValue',
+      () => expect(() => baz.typeVariableField, throwsUnsupportedError),
+    );
+
+    test(
```
srawlins, line 270: I'd like some more tests for `setDefaultDummyValue`. They could be added here or in another file like `mockito_test.dart`. They can be shaped just like this one, where `setDefaultDummyValue` is set, and mock method is called:

* `setDefaultDummyValue()` without a type argument
* `setDefaultDummyValue<X>()` with `typedef X = int`
* `setDefaultDummyValue<Foo<Bar>>()` (both Foo and Bar being imported)
* Maybe something showing subtypes don't work, so `setDefaultDummyValue<num>(7)` and then `baz.returnsTypeVariable<int>()` I think shouldn't work.

srawlins (changes requested):

zhanglix:
any update on this?

lib/src/mock.dart
```diff
@@ -1267,3 +1267,27 @@ extension ListOfVerificationResult on List<VerificationResult> {
   /// [verifyInOrder].
   List<List<dynamic>> get captured => [...map((result) => result.captured)];
 }
+
+final _defaultDummyValueFor = <Type, dynamic>{
```
srawlins, line 1271: We should use `Object?` instead of `dynamic`. Nothing here requires dynamic dispatch.

test/builder/auto_mocks_test.dart
```diff
@@ -3037,79 +3037,55 @@ void main() {
...
 
   test(
-      'throws when GenerateMocks is given a class with a getter with a '
-      'non-nullable class-declared type variable type', () async {
-    _expectBuilderThrows(
-      assets: {
-        ...annotationsAsset,
-        ...simpleTestAsset,
-        'foo|lib/foo.dart': dedent('''
+      'GenerateMocks works given class with a getter with a non-nullable '
```
srawlins, line 3040: Instead of 'works', we can say something more specific, like

`'uses defaultDummyValue, given a class...'`

Same with the 'works' text in the tests below.

test/end2end/generated_mocks_test.dart
```diff
@@ -226,8 +232,29 @@ void main() {
...
+        'typeVariableField can be accessed after setting default '
+        'via setDefaultDummyValue', () {
+      var value = const Foo<int>();
+      setDefaultDummyValue<Foo<int>>(value);
+      when(baz.typeVariableField).thenReturn(value);
+      expect(baz.typeVariableField, value);
+    });
+
+    test('setDefaultDummyValue() without type wont make harm', () {
```
srawlins, line 254: instead of 'wont make harm', we could say something like 'does not throw'.

srawlins (changes requested):
Some drive-by comments.

srawlins:
@yanok I don't understand this comment:

> I'd love to also have a generated code that populates the map with created fake classes. Seems like a free win.

yanok:
I meant we could (and I think we actually should) generate does
```dart
setDefaultDummyValue<Foo>(FakeFoo());
```
for every `FakeFoo` we generate.

The tricky part is we would need to somehow call this generated code for it to be useful. I was thinking about generating something like

```dart
bool _bla_mocks_inited = false;
void _bla_mocks_init() {
  if (_bla_mocks_inited) return;
  setDefaultDummyValue<Foo>(FakeFoo());
  // ...
  _bla_mocks_inited = true;
}
```

and putting a call to `_bla_mocks_init()` to all generated mocks' constructors. That's super ugly but should work I think.

But I have a more generic thought about this approach with a map. If we are going to use it anyway, why do we use it as a last resort only? Why don't just scrap all the cases and keep only this case that does a map lookup? Of course, ensuring that everything is added into the map in the generated init() function. That will probably simplify the codegen 
dramatically.

But we could take it further: armed with the default value map, we could rewrite our matchers to return `T` instead of `Null`. Well, except for `any` and another one, which are getters and can't be polymorphic, we would need to either get polymorphic getters approved or change the API to use `any()` method instead. But imagine we could do that, we could stop doing the dreading argument type nullification then! Just recently I had to explain people the error messages they got because of that several times...

And if we also add a typed `Invocation<T>` on top of that, that opens a path to make it kinda useful without codegen at all: we will still recommend using codegen to populate the default value table, but if people would be fine with doing it manually, that will work too.

srawlins:
> Why don't just scrap all the cases and keep only this case that does a map lookup? Of course, ensuring that everything is added into the map in the generated init() function. That will probably simplify the codegen
dramatically.

👍 I like it.

> we would need to either get polymorphic getters approved or change the API to use any() method instead. But imagine we could do that, we could stop doing the dreaded argument type nullification then!

👍 I like it. Just need an incremental migration path.

> And if we also add a typed Invocation<T> on top of that, that opens a path to make it kinda useful without codegen at all: we will still recommend using codegen to populate the default value table, but if people would be fine with doing it manually, that will work too.

👍

yanok:
> > we would need to either get polymorphic getters approved or change the API to use any() method instead. But imagine we could do that, we could stop doing the dreaded argument type nullification then!
> 
> +1 I like it. Just need an incremental migration path.

actually there are some recent comments in the [generic getters proposal](https://github.com/dart-lang/language/issues/1622) and they seem to be less negative than before ;) So probably there is hope. Then we could just wait and keep the API.

zhanglix:
> But we could take it further: armed with the default value map, we could rewrite our matchers to return `T` instead of `Null`. Well, except for `any` and another one, which are getters and can't be polymorphic, we would need to either get polymorphic getters approved or change the API to use `any()` method instead. But imagine we could do that, we could stop doing the dreaded argument type nullification then! Just recently I had to explain people the error messages they got because of that several times...
> 

> And if we also add a typed `Invocation<T>` on top of that, that opens a path to make it kinda useful without codegen at all: we will still recommend using codegen to populate the default value table, but if people would be fine with doing it manually, that will work too.

These are really great ideas! I am eagerly looking forward to using mockito without codegen. @yanok 👍

zhanglix:
> Some drive-by comments.

@srawlins Thank you for the comments. All comments are addressed in latest commits.

test/builder/auto_mocks_test.dart
```diff
@@ -3037,79 +3037,55 @@ void main() {
...
-      message: contains(
-          "The property accessor 'Foo.f' features a non-nullable unknown "
-          'return type, and cannot be stubbed'),
+      _containsAllOf('returnValue: _i1.defaultDummyValue<T>()'),
     );
   });
 
   test(
-      'throws when GenerateMocks is given a class with a method with a '
+      'uses defaultDummyValue, given a a class with a method with a '
```
srawlins, line 3053: whoops, not "a a class"

lib/mockito.dart
```diff
@@ -51,4 +51,8 @@ export 'src/mock.dart'
         logInvocations,
         untilCalled,
         MissingStubError,
-        FakeUsedError;
+        FakeUsedError,
+
+        //to support non-nullable return types
```
srawlins, line 56: Nit: Space after `//` and format like a sentence. https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences

srawlins (changes requested):
This is great!

@yanok would you like to review?

zhanglix:
@srawlins all comments fixed. Can you review it again? Thanks.

drown0315:
can't wait to see it.😀

srawlins:
@yanok should be the final vote here.

zhanglix:
> @yanok should be the final vote here.

@yanok Could you review it ?

zhanglix:
Any updates?

srawlins:
Sorry, no updates.

yanok:
Slightly different implementation of the same functionality is merged now. Closing this PR. Thanks for your idea!

PR #597, opened 2023-01-15
Add `returnInOrder` to `Mock`

danielgomezrico:
Fix:
- https://github.com/dart-lang/mockito/issues/260

Allows you to add multiple return options and it will return them in order of calling, like:

```dart
when(calculator.sum(any, any)).thenReturnInOrder([2, 4]);
```

Will cause that the code will return each expected result on every call:
```dart
calculator.sum(any, any); // Returns 2
calculator.sum(any, any); // Returns 4
calculator.sum(any, any); // Exception
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Require Dart >= 2.19.0.
 * Require analyzer 5.11.0.
 * Fail to generate mocks of `sealed`/`base`/`final` classes.
+* Add new `thenReturnInOrder` method to mock multiple calls to a single method in order.
 
 ## 5.4.0
 
--- a/README.md
+++ b/README.md
@@ -103,6 +103,12 @@ var responses = ["Purr", "Meow"];
 when(cat.sound()).thenAnswer((_) => responses.removeAt(0));
 expect(cat.sound(), "Purr");
 expect(cat.sound(), "Meow");
+
+// We can stub a method with multiple calls that happened in a particular order.
+when(cat.sound()).thenReturnInOrder(["Purr", "Meow"]);
+expect(cat.sound(), "Purr");
+expect(cat.sound(), "Meow");
+expect(() => cat.sound(), throwsA(isA<StateError>()));
 ```
 
 The [`when`], [`thenReturn`], [`thenAnswer`], and [`thenThrow`] APIs provide a
--- a/lib/src/mock.dart
+++ b/lib/src/mock.dart
@@ -18,6 +18,7 @@
 // ignore_for_file: prefer_void_to_null
 
 import 'dart:async';
+import 'dart:collection';
 
 import 'package:meta/meta.dart';
 import 'package:mockito/src/call_pair.dart';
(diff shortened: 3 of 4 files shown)


danielgomezrico:
@kevmoo or @srawlins ping 🥇

srawlins:
@yanok would you like to review?

lib/src/mock.dart
```diff
@@ -499,17 +503,21 @@ class PostExpectation<T> {
...
 
+  void thenReturnInOrder(List<T> expects) {
+    if (expects.isEmpty) {
+      throw ArgumentError('thenReturnInOrder expects should not be empty');
+    }
+
+    expects.forEach(_throwIfInvalid);
+
+    final answers = Queue.of(expects);
+    thenAnswer((_) => answers.removeFirst());
```
yanok, line 518: Should we check `answers` inside the closure and report a more specific error than "`Queue.removeFirst()` called on empty queue" if a method was called more times than the length of the given list?

test/mockito_test.dart
```diff
@@ -239,6 +246,13 @@ void main() {
           throwsArgumentError);
     });
 
+    test('thenReturn throws if order', () {
```
yanok, line 249: Test description should be more precise. "thenReturnInOrder throws on empty list input" or something like that.

yanok (changes requested):

lib/src/mock.dart
danielgomezrico: Thanks for the review, I made an update with your suggestion, do you see that the thrown error is fine?

test/mockito_test.dart
danielgomezrico: Thanks for the suggestion, I made the update, WDYT?

danielgomezrico:
@yanok thanks for the review, I made some changes, WDYT now?

test/mockito_test.dart
```diff
@@ -168,6 +168,25 @@ void main() {
       expect(mock.getter, equals('A'));
     });
 
+    test('throws an exception if not enought answers were provided', () {
+      when(mock.methodWithNormalArgs(any)).thenReturnInOrder(['One', 'Two']);
+      run() => mock.methodWithNormalArgs(100);
```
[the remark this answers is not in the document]
danielgomezrico: @yanok done! Is it fixed?

I'm curious, where did you see that Linter? I can't see it on my side.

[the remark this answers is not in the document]
danielgomezrico: @yanok where is the documentation that you refer?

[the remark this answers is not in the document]
yanok: The linter fires when I try to import the code into Google's internal monorepo. Apparently it's not enabled in `pubspec.yaml` for Mockito, but it is enabled internally.

Re: docs:
1. please add a doc-comment to the new method.
2. please add a CHANGELOG entry.
3. (preferrably) add an example usage to the `README.md`. https://github.com/dart-lang/mockito#how-about-some-stubbing is probably the right place, or create a small subsection just for `returnInOrder`. Please also cover the case when we run out of provided answers.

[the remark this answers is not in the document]
danielgomezrico: @yanok thanks for the explanation, I added the docs. I'm sorry if my English is not the best, let me know if you see something else that could improve 👍🏻

README.md
```diff
@@ -103,6 +103,12 @@ var responses = ["Purr", "Meow"];
 when(cat.sound()).thenAnswer((_) => responses.removeAt(0));
 expect(cat.sound(), "Purr");
 expect(cat.sound(), "Meow");
+
+// We can stub a method with multiple calls that happened in a particular order
```
yanok, line 107: Nit: comment sentences should end with a dot.

yanok (commented):
Please run `dart format` on it, the check is complaining. Other than that, LG.

README.md
danielgomezrico: Thanks! I fixed the comment and the format now

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
   `generate_for` argument to include files with mock generating annotations.
 * Require Dart >= 2.17.0.
 * Require analyzer 5.2.0.
+* Add new `thenReturnInOrder` method to mock multiple calls to a single method in order.
```
yanok, line 20: Ouch, that's a wrong place, we are not adding this new method to the already released version :) Please move up to the -wip section.

yanok: Could you also please do me a favor and squash it all in one commit with the informative commit message (like the one you used for this PR).

PR #685, opened 2023-08-10
Make MockBuilder support build_extensions option.

LuisDuarte1:
This PR changes the behaviour of the `buildMocks` factory, in order to support the `build_extensions` option. 

This is useful in order to change the destination directory of the mocks, allowing for more customizability as in [this article](https://simonbinder.eu/posts/build_directory_moves/). 

This change is a bit opinionated, because it forces the end-user to have a output filename pattern that ends with `.mocks.dart` in order to be consistent with the already generated files from mockito. It also forces the user, to have an input pattern ending with `.dart`.

Closes https://github.com/dart-lang/mockito/issues/545

---

<details>

- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).

</details>

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -10,8 +10,13 @@ targets:
           # to your desired path. The default behaviour is to the .mocks.dart file to be in the same
           # directory as the source .dart file. As seen below this is customizable, but the generated
           # file must always end in `.mocks.dart`.
+          #
+          # If you specify custom build_extensions you MUST ensure that they cover all input files
           build_extensions:
             '^example/build_extensions/{{}}.dart' : 'example/build_extensions/mocks/{{}}.mocks.dart'
+            '^example/example.dart' : 'example/example.mocks.dart'
+            '^example/iss/{{}}.dart' : 'example/iss/{{}}.mocks.dart'
+            '^test/end2end/{{}}.dart' : 'test/end2end/{{}}.mocks.dart'
 
 builders:
   mockBuilder:
--- a/build.yaml
+++ b/build.yaml
@@ -5,6 +5,13 @@ targets:
         generate_for:
           - example/**.dart
           - test/end2end/*.dart
+        options:
+          # build_extensions takes a source pattern and if it matches it will transform the output
+          # to your desired path. The default behaviour is to the .mocks.dart file to be in the same
+          # directory as the source .dart file. As seen below this is customizable, but the generated
+          # file must always end in `.mocks.dart`.
+          build_extensions:
+            '^example/build_extensions/{{}}.dart' : 'example/build_extensions/mocks/{{}}.mocks.dart'
 
 builders:
   mockBuilder:
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -63,18 +63,40 @@ import 'package:source_gen/source_gen.dart';
 /// 'foo.mocks.dart' will be created.
 class MockBuilder implements Builder {
   @override
+  final Map<String, List<String>> buildExtensions = {
+    '.dart': ['.mocks.dart']
+  };
+
+  MockBuilder({Map<String, List<String>>? buildExtensions}) {
+    this.buildExtensions.addAll(buildExtensions ?? {});
+  }
+
+  @override
   Future<void> build(BuildStep buildStep) async {
     if (!await buildStep.resolver.isLibrary(buildStep.inputId)) return;
     final entryLib = await buildStep.inputLibrary;
     final sourceLibIsNonNullable = entryLib.isNonNullableByDefault;
-    final mockLibraryAsset = buildStep.inputId.changeExtension('.mocks.dart');
+
+    // While it can be acceptable that we get more than 2 allowedOutputs,
+    // because it's the general one and the user defined one. Having
+    // more, means that user has conflicting patterns so we should throw.
+    if (buildStep.allowedOutputs.length > 2) {
+      throw ArgumentError('Build_extensions has conflicting outputs on file '
+          '`${buildStep.inputId.path}`, it usually caused by missconfiguration '
+          'on your `build.yaml` file');
+    }
+    // if not single, we always choose the user defined one.
+    final mockLibraryAsset = buildStep.allowedOutputs.singleOrNull ??
+        buildStep.allowedOutputs
+            .where((element) =>
+                element != buildStep.inputId.changeExtension('.mocks.dart'))
+            .single;
     final inheritanceManager = InheritanceManager3();
     final mockTargetGatherer =
         _MockTargetGatherer(entryLib, inheritanceManager);
 
-    final entryAssetId = await buildStep.resolver.assetIdForElement(entryLib);
     final assetUris = await _resolveAssetUris(buildStep.resolver,
-        mockTargetGatherer._mockTargets, entryAssetId.path, entryLib);
+        mockTargetGatherer._mockTargets, mockLibraryAsset.path, entryLib);
 
     final mockLibraryInfo = _MockLibraryInfo(mockTargetGatherer._mockTargets,
         assetUris: assetUris,
@@ -240,11 +262,6 @@ $rawOutput
     }
     return element.library!;
   }
-
-  @override
-  final buildExtensions = const {
-    '.dart': ['.mocks.dart']
-  };
 }
 
 /// An [Element] visitor which collects the elements of all of the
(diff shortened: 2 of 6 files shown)


yanok (changes requested):
LGTM, but we need:
1. Some documentation update to mention this feature, maybe even a FAQ entry;
2. Maybe add `build_extensions` to `example/` project?

yanok:
Corr: actually `example/` is not a separate project, so maybe just change the top-level `build.yaml` file to show how it works?

LuisDuarte1:
> LGTM, but we need:
> 
> 1. Some documentation update to mention this feature, maybe even a FAQ entry;
> 2. Maybe add `build_extensions` to `example/` project?

I agree so I've added them.

Also, I found some edge cases with my implementation so I took care of them, I'll list them: 
- Conflicting outputs, the user can supply multiple `build_extensions` and the builder will throw if it finds a conflicting output
- If the user provided a build_extension that doesn't match all .dart files that can be processed, the non-matched files would be ignored
- Relative includes in the `.mocks.dart` weren't being properly resolved if the output was in a different directory.

lib/src/builder.dart
```diff
@@ -62,19 +62,41 @@ import 'package:source_gen/source_gen.dart';
...
+  }
+
   @override
   Future<void> build(BuildStep buildStep) async {
     if (!await buildStep.resolver.isLibrary(buildStep.inputId)) return;
     final entryLib = await buildStep.inputLibrary;
     final sourceLibIsNonNullable = entryLib.isNonNullableByDefault;
-    final mockLibraryAsset = buildStep.inputId.changeExtension('.mocks.dart');
+
+    //While it can be acceptable that we get more than 2 allowedOutputs,
```
yanok, line 80: nit: please add spaces between `//` and text.

yanok (approved):
Thanks!

build.yaml
```diff
@@ -5,6 +5,13 @@ targets:
         generate_for:
           - example/**.dart
           - test/end2end/*.dart
+        options:
+          # build_extensions takes a source pattern and if it matches it will transform the output
+          # to your desired path. The default behaviour is to the .mocks.dart file to be in the same
+          # directory as the source .dart file. As seen below this is customizable, but the generated
+          # file must always end in `.mocks.dart`.
```
yanok, line 12:
```suggestion
          # file must always end in `.mocks.dart`.
```

build.yaml
```diff
@@ -5,6 +5,13 @@ targets:
         generate_for:
           - example/**.dart
           - test/end2end/*.dart
+        options:
+          # build_extensions takes a source pattern and if it matches it will transform the output
+          # to your desired path. The default behaviour is to the .mocks.dart file to be in the same
+          # directory as the source .dart file. As seen below this is customizable, but the generated
+          # file must always end in `.mocks.dart`. 
+          build_extensions:
+            '^example/build_extensions/{{}}.dart' : 'example/build_extensions/mocks/{{}}.mocks.dart'
```
yanok, line 14:
```suggestion
            '^example/build_extensions/{{}}.dart' : 'example/build_extensions/mocks/{{}}.mocks.dart'
```

lib/src/builder.dart
```diff
@@ -62,19 +62,41 @@ import 'package:source_gen/source_gen.dart';
 /// will produce a "'.mocks.dart' file with such mocks. In this example,
 /// 'foo.mocks.dart' will be created.
 class MockBuilder implements Builder {
+  @override
+  final Map<String, List<String>> buildExtensions = {
+    ".dart": [".mocks.dart"]
```
yanok, line 67:
```suggestion
    '.dart': ['.mocks.dart']
```

lib/src/builder.dart
```diff
@@ -62,19 +62,41 @@ import 'package:source_gen/source_gen.dart';
...
+    if (buildStep.allowedOutputs.length > 2) {
+      throw ArgumentError('Build_extensions has conflicting outputs on file '
+          '`${buildStep.inputId.path}`, it usually caused by missconfiguration '
+          'on your `build.yaml` file');
+    }
+    // if not single, we always choose the user defined one.
+    final mockLibraryAsset = buildStep.allowedOutputs.singleOrNull ??
+        buildStep.allowedOutputs
+            .where((element) =>
+                element != buildStep.inputId.changeExtension(".mocks.dart"))
```
yanok, line 92:
```suggestion
                element != buildStep.inputId.changeExtension('.mocks.dart'))
```

yanok (changes requested):
Could you please accept my changes and squash all commits into one?

natebosch:
> and squash all commits into one?

We should be able to squash for the internal commit - generally it's easier to avoid force-pushing over shared history in a PR.

natebosch:
> should I revert it then?

No need - we can work with it either way. I was just hoping to inform on best practices since I frequently use shared PR history when reviewing in `dart-lang` reps 😄

lib/src/builder.dart
```diff
@@ -62,19 +62,41 @@ import 'package:source_gen/source_gen.dart';
...
+
   @override
   Future<void> build(BuildStep buildStep) async {
     if (!await buildStep.resolver.isLibrary(buildStep.inputId)) return;
     final entryLib = await buildStep.inputLibrary;
     final sourceLibIsNonNullable = entryLib.isNonNullableByDefault;
-    final mockLibraryAsset = buildStep.inputId.changeExtension('.mocks.dart');
+
+    // While it can be acceptable that we get more than 2 allowedOutputs,
+    // because it's the general one and the user defined one. Having
```
natebosch, line 81: I don't understand this - user defined configuration for the `.dart` input will override the default. I would expect that the allowed outputs should always have a single value.

lib/src/builder.dart
```diff
@@ -2304,7 +2321,29 @@ class _AvoidConflictsAllocator implements Allocator {
...
+  final result = <String, List<String>>{};
+  for (final entry in buildExtensions.entries) {
+    final input = entry.key;
+    final output = entry.value;
+    if (input is! String || !input.endsWith('.dart')) {
+      throw ArgumentError('Invalid key in build_extensions `$input`, it '
+          'should be a string ending with `.dart`');
+    }
+    if (output is! String || !output.endsWith('.mocks.dart')) {
+      throw ArgumentError('Invalid key in build_extensions `$output`, it '
```
natebosch, line 2340:
```suggestion
      throw ArgumentError('Invalid value in build_extensions `$output`, it '
```

lib/src/builder.dart
natebosch: Oh I see, the examples are using a `^test` in the build extensions to substitute for a `generate_for`. cc @jakemac53 - did we predict this would happen when we enhanced the build extension syntax?


Maybe it would be better to _only_ do a full override and not do a merge.

```dart
  @override
  final Map<String, List<String>> buildExtensions;

  MockBuilder({Map<String, List<String>>? buildExtensions})
      : buildExtensions = buildExtensions ??
            const {
              '.dart': ['.mocks.dart'],
            };
```

LuisDuarte1: Having the commit history would be helpful now ahah, but that was my first implementation, but there's a reason I changed it into a merge. I've discovered this when making the example for this config, `example/build_extensions`, when doing the full override and only having an input pattern that only **partially** covers all `.dart` files, the unmatched `.dart` files wouldn't be processed, even though there's a `generate_for` for those files. So my fix was to merge the user rules with the generic one  ( `.dart`=>`.mocks.dart`)

yanok:
> We should be able to squash for the internal commit - generally it's easier to avoid force-pushing over shared history in a PR.

Well, actually Copybara already does the squash, so it's mostly about getting a nicer CL description, as now it will list all the commits. We should probably just fix our Copybara config to copy the PR description into CL.

yanok:
We need to wait for the resolution of the discussion above. Also please convert the added double-quotes in the test file into single quotes: we have a lint complaining about that internally.

lib/src/builder.dart
natebosch: > when doing the full override and only having an input pattern that only **partially** covers all `.dart` files, the unmatched `.dart` files wouldn't be processed, even though there's a `generate_for` for those files.

I think not processing them is preferable to silently falling back on potentially unexpected behavior. I think it is reasonable to expect the overridden config to cover all inputs.

LuisDuarte1: > I think not processing them is preferable to silently falling back on potentially unexpected behavior. I think it is reasonable to expect the overridden config to cover all inputs.

I can understand your perspective, in this case it's best to be consistent and not introduce any unexpected behavior, even though I think my approach is more ergonomic for developers. I'll try to change it then.

lib/src/builder.dart
```diff
@@ -62,19 +62,33 @@ import 'package:source_gen/source_gen.dart';
...
     final entryLib = await buildStep.inputLibrary;
     final sourceLibIsNonNullable = entryLib.isNonNullableByDefault;
-    final mockLibraryAsset = buildStep.inputId.changeExtension('.mocks.dart');
+
+    if (buildStep.allowedOutputs.length > 1) {
+      throw ArgumentError('Build_extensions has conflicting outputs on file '
+          '`${buildStep.inputId.path}`, it usually caused by missconfiguration '
+          'on your `build.yaml` file');
+    }
+    final mockLibraryAsset = buildStep.allowedOutputs.single;
```
natebosch, line 84: [nit] The `single` could fail if the list was empty. How about

```suggestion
    final mockLibraryAsset = buildStep.allowedOutputs.singleOrNull;
    if (mockLibraryAsset == null) {
      throw ArgumentError('Build_extensions has missing or conflicting outputs for '
          '`${buildStep.inputId.path}`, this is usually caused by a misconfigured '
          'build extension override in `build.yaml`');
    }
```

natebosch (approved):

lib/src/builder.dart
LuisDuarte1: Nice catch!

yanok (approved):

yanok:
Now the formatter complains :) Could you please re-format it?

LuisDuarte1:
> Now the formatter complains :) Could you please re-format it?

Done. It seems that Dart 3.0 has a different formatting than Dart 2.19, a quick setup in github codespaces allowed me to format the files in the CI version :sweat_smile:.

yanok:
> Done. It seems that Dart 3.0 has a different formatting than Dart 2.19, a quick setup in github codespaces allowed me to format the files in the CI version 😅.

Arghh, I see. Yes, I think we did change the formatter some time ago. Damn, then we should have just bumped the required SDK version I guess... I wonder if now formatter will start complaining internally, since we use more or less HEAD version... But let's wait.

yanok:
Nope, it worked fine, thanks!

PR #686, opened 2023-08-15
Fix analysis errors in example directory

parlough:


example/example.dart
```diff
@@ -35,7 +33,7 @@ abstract class Callbacks {
   Cat,
   Callbacks,
 ], customMocks: [
-  MockSpec<Cat>(as: #MockCatRelaxed, returnNullOnMissingStub: true),
+  MockSpec<Cat>(as: #MockCatRelaxed),
```
yanok, line 36: We don't really show the usage of it here, but this won't make it relaxed: for `GenerateMocks` the default is to throw. If we want to show a relaxed version, we have to pass `onMissingStub: OnMissingStub.returnDefault`.

parlough: Thanks for the call out. I specified the `OnMissingStub.returnDefault` and added an example of the resulting behavior. Let me know what you think.

yanok (approved):

PR #700, opened 2023-10-04
Use 3.0.0 as stable SDK in CI

yanok:


yanok:
Formatting needs to be fixed, I'll do it in a separate CL.

.github/workflows/test-package.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608
       - uses: dart-lang/setup-dart@8a4b97ea2017cc079571daec46542f76189836b1
         with:
-          sdk: 2.19.0
+          sdk: 3.1.3
```
natebosch, line 27: Can we use `3.0.0` to match the SDK constraint?

yanok: Changed it to 3.0.0. But it seems now I need to roll back the formatting change.

srawlins (approved):

PR #719, opened 2023-11-16
Update to the latest lints, require Dart 3.1

kevmoo:


yanok (approved):
I'm not super sure, but I think we usually drop -wip in a separate commit. Please check with nbosch@

And if you end up bumping the version here, there is also something like version_test.dart that needs to be updated.

PR #728, opened 2023-12-20
Enable MockSpec superclasses

oznecniV97:
In this PR I add a check on mockSpec type to be sure that is an effective MockSpec and not an extension of it. In this way we could use a MockSpec superclass to use default constructor parameters or added values.

---

<details>

- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).

</details>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ coverage/
 
 # ignore generated mocks
 **/*.mocks.dart
+
+# ide folders
+.idea
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -574,6 +574,11 @@ class _MockTargetGatherer {
       int index,
       List<ast.CollectionElement> mockSpecAsts,
       {bool nice = false}) {
+    // need to get the parent until type name is not MockSpec to allow inheritance
+    while(mockSpec.type?.element?.name != "MockSpec") {
+      mockSpec = mockSpec.getField("(super)")!;
+    }
+
     final mockSpecType = mockSpec.type as analyzer.InterfaceType;
     assert(mockSpecType.typeArguments.length == 1);
     final mockType = _mockType(mockSpecAsts[index]);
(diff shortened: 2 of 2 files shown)


srawlins:
Thanks for the PR. Please file a bug describing what this is fixing. We also need tests showing what would fail without the changes in `lib/`.

oznecniV97:
Issue opened with example test: https://github.com/dart-lang/mockito/issues/729

oznecniV97:
Any update on this thread? There's some approvers?

yanok:
Sorry for the delay: this project has only a single mantainer, who has more urgent tasks most of the time.

Re: discussion on the bug: It's a pity we can't have a `const` function, I see now, why you want to extend `MockSpec`.

I still don't quite like your solution though: surely, it works for your case with `fallbackGenerators` passed to the super constructor. But what if someone would do:
```dart
class MyMockSpec<T> extends MockSpec<T> {
  const MyMockSpec();
  @override
  final fallbackGenerators = const {#x: fallback};
}
```

Your code would simply ignore it.

Furthermore, it's technically possible now to _implement_ `MockSpec` instead of _extending_ it, so your code will never find `MockSpec` in superclasses and crash. Well, arguably that not worse than what we already have. We could also forbid this by making `MockSpec` `base`.

To cover all cases we need to call `getField` while going down the inheritance chain (I'm actually surprised why Analyzer can't do that for us). The only thing there you indeed want to get to `MockSpec` itself, is getting the type parameter, to cover weird cases like
```dart
class MyMockSpecForFoo extends MockSpec<Foo> {}
class MyMockSpecForListT<T> extends MockSpec<List<T>> {}
```

Would you be up for covering all of this? Especially provided that I want to do the next major release by the end of the end, hoping to be able to drop the fallback generators concept entirely. Privided there is no Analyzer API to help us here, personally I kinda feel the the code complication doesn't worth the improvement. But if you still want to follow this route, please also add cases cases covering various way to subtype `MockSpec`.

As a workaround you can define your fallback generators as a top-level `const` and pass it to every `MockSpec`. Not quite what you wanted to achieve, but still probably better than repeating the map every time.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 5.4.5-wip
+
+* Allow usage of GetMock superclasses.
```
yanok, line 3:
```suggestion
* Allow usage of MockSpec subclasses.
```

PR #754, opened 2024-06-06
Provide better documentation for `provideDummy`/`provideDummyBuilder`

jamesderlin:
Mostly fixes https://github.com/dart-lang/mockito/issues/697.

I'm still not quite satisfied with the documentation for `provideDummyBuilder` since the `DummyBuilder` `typedef` remains undocumented, but I can't write documentation for that because I don't understand its arguments.

---


jamesderlin:
@srawlins Could you take a look when you have a chance?

FAQ.md
```diff
@@ -203,3 +203,71 @@ targets:
 ```
 
 Also, you can also check out the example configuration in the Mockito repository. 
+
+
+## How do I mock a `sealed` class?
```
srawlins, line 208: This section discusses how to mock a class that requires instances of sealed classes, rather that mocking sealed classes.

FAQ.md
```diff
@@ -203,3 +203,71 @@ targets:
...
+
+  Bar.zero() : value = 0;
+}
+
+class Foo {
+  Bar someMethod(int value) => Bar(value);
+}
+```
+
+and now you want to mock `Foo`. The generated implementation for `MockFoo`
```
srawlins, line 226: I think this is a bit of a leap from wanting to mock `Foo` to having a generated implementation. Maybe something like, "A mock implementation of `Foo`, generated by `@GenerateNiceMocks([MockSpec<Foo>()])`, needs to ..."

FAQ.md
```diff
@@ -203,3 +203,71 @@ targets:
...
+without understanding the semantics of `Bar`'s constructors. That is, which
+`Bar` constructor should be called and with what arguments? To avoid this,
+Mockito instead generates its own, fake implementation of `Bar` that it does
+know how to construct, something like:
+
+```dart
+class _FakeBar extends Fake implements Bar {}
+```
+
+And then the generated implementation of `MockFoo` can have its `someMethod`
```
srawlins, line 238: It's not really an "implementation" of `MockFoo`; you could say the generated _declaration_ of `MockFoo` or maybe the generated mock implementation of `Foo`.

lib/src/dummies.dart
```diff
@@ -156,7 +157,8 @@ T dummyValue<T>(Object parent, Invocation invocation) {
   throw MissingDummyValueError(T);
 }
 
-/// Provide a builder for that could create a dummy value of type `T`.
+/// Specifies a builder to create a dummy value of type `T`.
```
srawlins, line 160: Thanks :D

srawlins (changes requested):
Looks great! A few comments

FAQ.md
jamesderlin: Oops.  Heh, I even made a similar correction to a StackOverflow question about this.

I also updated the URLs referring to this section. I feel that the new heading is too long and is more brittle (and I don't think GitHub Markdown supports generating custom slugs?), so I've changed the URLs to be a bit less specific.

FAQ.md
jamesderlin: I do like that better.  Fixed.

FAQ.md
jamesderlin: Argh, yeah, I didn't mean "implementation" in the Dart `implements` sense.  I will just remove "the generated implementation of".

srawlins (approved):

PR #776, opened 2024-10-29
Resolve for typedefs/aliases and records in `typeArguments` & resolve duplicate typedefs

dickermoshe:
Closes: https://github.com/dart-lang/mockito/issues/775 
Closes: https://github.com/dart-lang/mockito/issues/777 
Closes: https://github.com/dart-lang/mockito/issues/778

The fix for these issues is closely related so I've done both of them in a single PR.

- Mockito wasn't checking for typedefs/aliases in `typeArguments`
- it also wasn't looking into records on type args. 
- Typedefs which had the same underlying type were ignored even if the typedef had a different name

These issues cause the dreaded `is missing from the asset URI mapping` error.

See the related issue for a repro of each of these. I've also added regression tests for all of these.

Looking forward to a swift review!

---

<details>

- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-change).

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 * Require dart_style >= 2.3.7, so that the current Dart language version can be
   passed to `DartFormatter`.
 * Add topics to `pubspec.yaml`.
-* Fix a bug where typedef-aliases in type arguments were not correctly
+* Fix a bug where type aliases in type arguments were not correctly
   resolved.
 * Fix a bug where record types were not correctly resolved.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 * Require dart_style >= 2.3.7, so that the current Dart language version can be
   passed to `DartFormatter`.
 * Add topics to `pubspec.yaml`.
+* Fix a bug where typedef-aliases in type arguments were not correctly
+  resolved.
+* Fix a bug where record types were not correctly resolved.
 
 ## 5.4.4
 
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -173,6 +173,10 @@ $rawOutput
     void addTypesFrom(analyzer.DartType type) {
       // Prevent infinite recursion.
       if (seenTypes.contains(type)) {
+        if (type.alias != null) {
+          // To check for duplicate typdefs that have different names
+          type.alias!.element.accept(typeVisitor);
+        }
         return;
       }
       seenTypes.add(type);
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -167,25 +167,28 @@ $rawOutput
     // `Future`, which is needed when overriding some methods which return
     // `FutureOr`.
     final typeVisitor = _TypeVisitor(entryLib.typeProvider.futureDynamicType);
-    final seenTypes = <analyzer.InterfaceType>{};
+    final seenTypes = <analyzer.DartType>{};
     final librariesWithTypes = <LibraryElement>{};
 
-    void addTypesFrom(analyzer.InterfaceType type) {
+    void addTypesFrom(analyzer.DartType type) {
       // Prevent infinite recursion.
       if (seenTypes.contains(type)) {
         return;
       }
       seenTypes.add(type);
-      librariesWithTypes.add(type.element.library);
-      type.element.accept(typeVisitor);
-      if (type.alias != null) type.alias!.element.accept(typeVisitor);
-      // For a type like `Foo<Bar>`, add the `Bar`.
-      type.typeArguments
-          .whereType<analyzer.InterfaceType>()
-          .forEach(addTypesFrom);
-      // For a type like `Foo extends Bar<Baz>`, add the `Baz`.
-      for (final supertype in type.allSupertypes) {
-        addTypesFrom(supertype);
+      librariesWithTypes.addAll([
+        if (type.element?.library != null) type.element!.library!,
+        if (type.alias?.element.library != null) type.alias!.element.library,
+      ]);
+      type.element?.accept(typeVisitor);
+      type.alias?.element.accept(typeVisitor);
+      switch (type) {
+        case analyzer.InterfaceType interface:
+          interface.typeArguments.forEach(addTypesFrom);
+          interface.allSupertypes.forEach(addTypesFrom);
+        case analyzer.RecordType record:
+          record.positionalTypes.forEach(addTypesFrom);
+          record.namedTypes.map((e) => e.type).forEach(addTypesFrom);
       }
     }
 
--- a/lib/src/builder.dart
+++ b/lib/src/builder.dart
@@ -180,10 +180,14 @@ $rawOutput
         return;
       }
       seenTypes.add(type);
-      librariesWithTypes.addAll([
-        if (type.element?.library != null) type.element!.library!,
-        if (type.alias?.element.library != null) type.alias!.element.library,
-      ]);
+
+      if (type.element?.library case var library?) {
+        librariesWithTypes.add(library);
+      }
+      if (type.alias?.element.library case var library?) {
+        librariesWithTypes.add(library);
+      }
+
       type.element?.accept(typeVisitor);
       type.alias?.element.accept(typeVisitor);
       switch (type) {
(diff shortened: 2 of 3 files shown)


srawlins:
CC @mosuem

dickermoshe:
It seems that the bulk of this code is rely on `InterfaceType` as the only interface object, while records are interfaces themselves.
It seems that the `addTypesFrom` function, which until now only took a `InterfaceType` will now have to take records too.

dickermoshe:
Could this be cherrypicked into a 5.4.5 release?

dickermoshe:
CI is broken on the entire repo as dart stable does not support `analyzer 6.9`. Tests pass locally on my machine.

dickermoshe:
@mosuem 
I know people hate they get this, but.... Ping

mosuem:
I don't mind at all, but I am also not a maintainer of this repo and maybe not the right person to ping ;) I just stumbled across the same CI error in https://github.com/dart-lang/test/pull/2293

mosuem:
Maybe @srawlins wants to take a look (I hope he you don't mind the ping). I am fine refiling the PR to move to `dart-lang/test` and merging this first.

test/builder/auto_mocks_test.dart
```diff
@@ -3620,6 +3678,54 @@ void main() {
               contains('returnValue: _i3.Future<(int, {_i2.Bar bar})>.value('),
               contains('bar: _FakeBar_0('))));
     });
+    test('are supported as typedefs', () async {
```
srawlins, line 3681: Hmm, not 'typedefs', I think you mean 'type arguments'?

test/builder/auto_mocks_test.dart
```diff
@@ -3620,6 +3678,54 @@ void main() {
...
+            import 'package:mockito/annotations.dart';
+            @GenerateMocks([Foo])
+            void main() {}
+          '''
+      });
+
+      expect(mocksContent, contains('class MockFoo extends _i1.Mock'));
+      expect(mocksContent, contains('implements _i2.Foo'));
+    });
+    test('are supported as nested typedefs', () async {
```
srawlins, line 3705: "nested type arguments"

test/builder/auto_mocks_test.dart
```diff
@@ -3553,6 +3553,64 @@ void main() {
         expect(mocksContent, contains('class MockBaz extends _i1.Mock'));
         expect(mocksContent, contains('implements _i2.Baz'));
       });
+
+      test('when its a type parameter of function which returns another type',
```
srawlins, line 3557: nit: "it's" instead of "its"

srawlins, line 3557: I'm not sure what you mean here by "a type parameter of function which returns another type." In this test case, the typedef appears to only be used as a _type argument_ in a class definition `class Foo extends BaseFoo<CreateBar>`.

test/builder/auto_mocks_test.dart
```diff
@@ -3553,6 +3553,64 @@ void main() {
...
+        expect(mocksContent, contains('implements _i2.Foo'));
+      });
+      test('when its a duplicate type parameter', () async {
+        final mocksContent = await buildWithNonNullable({
+          ...annotationsAsset,
+          'foo|lib/foo.dart': dedent(r'''
+            class Bar {}
+            typedef BarDef = int Function();
+            typedef BarDef2 = int Function();
+            class BaseFoo<T,P> {
```
srawlins, line 3593: nit: space between `T,` and `P`, as that is how the formatter would format it.

test/builder/auto_mocks_test.dart
```diff
@@ -3553,6 +3553,64 @@ void main() {
...
+          'foo|test/foo_test.dart': '''
+            import 'package:foo/foo.dart';
+            import 'package:mockito/annotations.dart';
+
+            @GenerateMocks([Foo])
+            void main() {}
+          '''
+        });
+
+        expect(mocksContent, contains('class MockFoo extends _i1.Mock'));
```
srawlins, line 3611: I don't see any assertions here related to `BarDef` or `BarDef2`.

test/builder/auto_mocks_test.dart
```diff
@@ -3553,6 +3553,64 @@ void main() {
...
+        expect(mocksContent, contains('implements _i2.Foo'));
+      });
+      test('when its a duplicate type parameter', () async {
```
srawlins, line 3586: Instead of "duplicate" type parameter, do you just mean some type parameter after the first parameter? You could s say something like "any type parameter."

lib/src/builder.dart
```diff
@@ -167,25 +167,32 @@ $rawOutput
...
-      if (type.alias != null) type.alias!.element.accept(typeVisitor);
-      // For a type like `Foo<Bar>`, add the `Bar`.
-      type.typeArguments
-          .whereType<analyzer.InterfaceType>()
-          .forEach(addTypesFrom);
-      // For a type like `Foo extends Bar<Baz>`, add the `Baz`.
-      for (final supertype in type.allSupertypes) {
-        addTypesFrom(supertype);
+      librariesWithTypes.addAll([
+        if (type.element?.library != null) type.element!.library!,
```
srawlins, line 184: I think this can read better if we hoist these properties out with `if case`, instead of `addAll` and `if`:

```dart
if (type.element?.library case var library?) {
  librariesWithTypes.add(library);
}
if (type.alias?element.library case var library?) {
  librariesWithTypes.add(library);
}
```

The `var library?` syntax is still kind of weird to me, but I'm trying to use it more. Basically "if that is non-null then extract it to a variable named `library`."

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
 * Require dart_style >= 2.3.7, so that the current Dart language version can be
   passed to `DartFormatter`.
 * Add topics to `pubspec.yaml`.
+* Fix a bug where typedef-aliases in type arguments were not correctly
```
srawlins, line 12: I think instead of "typedef-aliases," we generally call them "type aliases."

srawlins (changes requested):
Thanks so much! I left a few notes but generally the fix is gold 😁

srawlins:
@mosuem I love a friendly ping 😁 I am fine doing whatever procedure works best for you, migrating to dart-lang/test. I think a git commit can be wired up that attributes the fix to @dickermoshe still...

For external PRs like this, I think I've been able to just hand-copy the PR, once landed, into google3, and then re-export. But there might be a good copybara mechanism for that, at this point...

mosuem:
I'll have to check the copybara settings, but normally even g3 SoT repos should be able to accept Github PRs. We can probably copy the config from package:intl, then it should just work.

mosuem:
But please go ahead with merging this, any infra work can wait, features are much nicer ;)

CHANGELOG.md
dickermoshe: Oh yeah, where do you think I got it from? 😉😜
https://github.com/dart-lang/mockito/blob/f72791d3dd5cb2acc4f5c7da065ffe70d2ad1e0b/CHANGELOG.md?plain=1#L51-L60

I'll fix it anyway 😂

lib/src/builder.dart
dickermoshe: Agree. Big fan of the `case` in `if` checks too.
It's a shame it isn't a well known feature.

test/builder/auto_mocks_test.dart
dickermoshe: How about?
`typedef mocks are generated properly when it\'s a function which returns any type`

In this instance, `Bar` was being ignored becuase the return type of `CreateBar` want being resolved

test/builder/auto_mocks_test.dart
dickermoshe: before this pr, the above test would throw an exception that `BarDef2` is not found, so the above is sufficient.
I'll add a small test to if you want me too.
```dart
expect(mocksContent, contains('_i2.BarDef get t1'));
expect(mocksContent, contains('_i2.BarDef2 get t2'));
```

dickermoshe:
Thanks for the timely response. This is the coolest library ever, thank so much!
Fixed all the nits, lmk if there is anything else

test/builder/auto_mocks_test.dart
srawlins: That SG, thanks!

srawlins:
OK we should be good to go. Ship it!

srawlins (approved):

PR #824, opened 2025-08-14
Allow analyzer 8.

davidmorgan:


davidmorgan:
Looks like there are a few upgrades needed first...

PR #834, opened 2025-10-03
Add dummy values for `RegExp` and `MapEntry`.

lrhn:
Avoids generating mocks for `RegExp` and `MapEntry`. Both are value types and shouldn't be mocked.

`RegExp` will eventually become non-mockable, `MapEntry` already is.


srawlins (approved):
Thanks for all the tidying!

